PR #32, opened 2019-11-11
deployment: Add shutdown command

marclop:
## Description

Adds the deployment shutdown command with an interactive prompt
confirming the action which can be skipped by using the `--force` flag.

## How Has This Been Tested?

```console
$ dev-cli deployment shutdown a3eb4382ccca482ea15ad1232aabf2cc
This action will delete the specified deployment ID and its associated sub-resources, do you want to continue? [y/n]: y
{
  "id": "a3eb4382ccca482ea15ad1232aabf2cc",
  "name": "",
  "orphaned": {
    "apm": [],
    "appsearch": [],
    "elasticsearch": [
      {
        "dependents": [
          {
            "id": "7e44eee3e62f416a8512811bd6498d6e",
            "kind": "kibana"
          }
        ],
        "id": "a3eb4382ccca482ea15ad1232aabf2cc"
      }
    ],
    "kibana": []
  }
}

```

## Types of Changes

### DIFF
--- a/cmd/deployment/shutdown.go
+++ b/cmd/deployment/shutdown.go
@@ -30,7 +30,7 @@ import (
 var shutdownCmd = &cobra.Command{
 	Use:     "shutdown <deployment-id>",
 	Short:   "Shutdown's a platform deployment",
-	PreRunE: cobra.MinimumNArgs(1),
+	PreRunE: cmdutil.MinimumNArgsAndUUID(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		force, _ := cmd.Flags().GetBool("force")
 		var msg = "This action will delete the specified deployment ID and its associated sub-resources, do you want to continue? [y/n]: "
--- a/cmd/deployment/shutdown.go
+++ b/cmd/deployment/shutdown.go
@@ -1,0 +1,62 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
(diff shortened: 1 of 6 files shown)


cmd/deployment/shutdown.go
```diff
@@ -0,0 +1,62 @@
...
+
+	cmdutil "github.com/elastic/ecctl/cmd/util"
+	"github.com/elastic/ecctl/pkg/deployment"
+	"github.com/elastic/ecctl/pkg/ecctl"
+)
+
+var shutdownCmd = &cobra.Command{
+	Use:     "shutdown <deployment-id>",
+	Short:   "Shutdown's a platform deployment",
+	PreRunE: cobra.MinimumNArgs(1),
```
ppapapetrou76: should we use `MinimumNArgsAndUUID(1)` instead?

pkg/deployment/shutdown.go
```diff
@@ -0,0 +1,73 @@
...
+// Validate ensures the parameters are usable by Shutdown.
+func (params ShutdownParams) Validate() error {
+	var merr = new(multierror.Error)
+
+	if params.API == nil {
+		merr = multierror.Append(merr, util.ErrAPIReq)
+	}
+
+	if params.DeploymentID == "" {
+		merr = multierror.Append(merr, util.ErrDeploymentID)
```
ppapapetrou76: this error message is not fully aligned with the actual check.
I would check the check condition to `if len(params.DeploymentID) != 32` or re-phrase the error message

PR #36, opened 2019-11-13
deployment: Add create command

marclop:
## Description

Adds the deployment create command which enables the creation of
multi-resource deployments from a file definition in one API call. The
main features are two flag overrides (name and version) which allow the
file definition to be used more like a template.

Additionally, a `--request-id` flag is provided which enables setting an
idempotency token for the API calls. By default an auto-generated value
is sent to the API and printed to the Standard Error device so that
users can use the token in case the deployment creation fails.

```
$ dev-cli deployment create -f deployment_example.json --name marcdeploy --version=7.4.1
The deployment creation returned with an error, please use the displayed idempotency token
to recreate the deployment resources
Idempotency token: GMZPMRrcMYqHdmxjIQkHbdjnhPIeBElcwrHwzVlhGUSMXrEIzVXoBykSVRsKncNb
unknown error (status 500)
```

## Motivation and Context

There's a small bug in the return code which returns a 201 instead of a 200 so the whole API call returns an error via the Go autogenerated code. Will push a fix on the SDK shortly.

## Types of Changes

### DIFF
--- a/cmd/deployment/create.go
+++ b/cmd/deployment/create.go
@@ -27,6 +27,7 @@ import (
 	cmdutil "github.com/elastic/ecctl/cmd/util"
 	"github.com/elastic/ecctl/pkg/deployment"
 	"github.com/elastic/ecctl/pkg/ecctl"
+	"github.com/elastic/ecctl/pkg/util"
 )
 
 const createLong = `Creates a deployment from a file definition with an automatically generated idempotency token.
@@ -144,7 +145,7 @@ var createCmd = &cobra.Command{
 
 		reqID, _ := cmd.Flags().GetString("request-id")
 		if reqID == "" {
-			reqID = cmdutil.RandomString(64)
+			reqID = util.RandomString(64)
 		}
 
 		res, err := deployment.Create(deployment.CreateParams{
--- a/cmd/deployment/create.go
+++ b/cmd/deployment/create.go
@@ -45,14 +45,17 @@ $ cat deployment_example.json
                 "ref_id": "my-apm-instance",
                 "elasticsearch_cluster_ref_id": "my-es-cluster",
                 "plan": {
-                    "zone_count": 1,
                     "apm": {
                         "version": "6.8.4"
                     },
-                    "size":{
-                        "resource": "memory",
-                        "value": 512
-                    }
+                    "cluster_topology": [{
+                        "instance_configuration_id": "apm",
+                        "size": {
+                            "resource": "memory",
+                            "value": 512
+                        },
+                        "zone_count": 1
+                    }]
                 }
             }
         ],
--- a/cmd/deployment/create.go
+++ b/cmd/deployment/create.go
@@ -30,7 +30,7 @@ import (
 )
 
 const createLong = `Creates a deployment from a file definition with an automatically generated idempotency token.
-On creation failure, please use the displayed idempotency token to retry the cluster creation.
+On creation failure, please use the displayed idempotency token to retry the cluster creation with --request-id=<token>.
 
 Read more about the deployment definition in https://www.elastic.co/guide/en/cloud-enterprise/current/Deployment_-_CRUD.html`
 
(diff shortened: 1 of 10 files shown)


cmd/deployment/create.go
```diff
@@ -0,0 +1,169 @@
...
+
+		return ecctl.Get().Formatter.Format("", res)
+	},
+}
+
+func init() {
+	Command.AddCommand(createCmd)
+	createCmd.Flags().String("name", "", "Overrides the deployment name")
+	createCmd.Flags().String("version", "", "Overrides all thee deployment's resources to the specified version")
+	createCmd.Flags().String("request-id", "", "Optional idempotency token - if two create requests share the same request_id token (min size 32 characters, max 128) then only one deployment will be created, the second request will return the info of that deployment (in the same format described below, but with blanks for auth-related fields)")
```
ppapapetrou76: I assume users know how to generate this token? Should we explain somehow? Or point them to some online documentation? For instance I have no idea what is this and how to generate it :)

pkg/deployment/create.go
```diff
@@ -0,0 +1,159 @@
...
+func setOverrides(req *models.DeploymentCreateRequest, overrides *CreateOverrides) {
+	if req == nil || overrides == nil || req.Resources == nil {
+		return
+	}
+
+	if overrides.Name != "" {
+		req.Name = overrides.Name
+	}
+
+	for _, resource := range req.Resources.Apm {
```
ppapapetrou76: Can we do something with these 4 almost identical `for` loops? I'd prefer to see them as a function or something. wdyt?

cmd/deployment/create.go
```diff
@@ -0,0 +1,169 @@
...
+	PreRunE: cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		filename, _ := cmd.Flags().GetString("file")
+		var r models.DeploymentCreateRequest
+		if err := cmdutil.DecodeFile(filename, &r); err != nil {
+			return err
+		}
+
+		var region string
+		if ecctl.Get().Config.Region == "" {
```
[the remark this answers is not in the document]
marclop: In the ECE world, there aren't regions, necessarily, at least there's no concept of federation or multiple regions in an ECE installation, as such, many if not all users won't necessarily be aware that the _default_ region is `ece-region` and thus, not have the `region` global configuration parameter set to any value, meaning empty.

If that is the case, I'm automatically making it `ece-region` as a default, and that will get also populated down the line to the deployment definition for each of the deployment's defined resources.

pkg/deployment/create.go
marclop: unfortunately as almost identical as they are, they're not, and even if the underlying operations are the same, there's different fields and they're of different types. It sucks, but this is what we end up with.

I'm not sure I understand what you mean by ` I'd prefer to see them as a function or something`

cmd/deployment/create.go
marclop: I can change the wording on this flag, but an explanation on the indempotency tokens can be found in line :32-34 and here: https://github.com/elastic/ecctl/pull/36/files#diff-437f06621adf011db767be2048d80d9eR7-R10

marclop: Let me know if f8842ac addresses your concerns

pkg/deployment/create.go
marclop: Do you mean writing a private function for each case?

cmd/deployment/create.go
marclop: Also th PR description has some info about it

ppapapetrou76: Yeah now it's super clear! 🙇 👍

cmd/deployment/create.go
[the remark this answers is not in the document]
ppapapetrou76: Yeah I understand that. My question is why are not doing anything similar [here](https://github.com/elastic/ecctl/blob/master/cmd/deployment/delete.go#L33) for instance? I'm pointing to this particular command because it's on the same API endpoint and was implemented recently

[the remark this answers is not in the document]
ppapapetrou76: I mean if we want to use a default region this totally fine, but I think we need to be consistent across all commands or at least keep it on all new commands implemented from now on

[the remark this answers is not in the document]
ppapapetrou76: We chatted on slack and I was confused, assuming that `region` is needed as part of every API call which is not true at all. So my previous comments are note valid. Thanks @marclop for clearing this out !

pkg/deployment/create.go
ppapapetrou76: Yeah I was thinking something like `processResource(res Resource)` but since they are completely different structures , meh.

We could add a wrapper interface on top of them by providing some functions but this is over-engineering.

cmd/util/random.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"math/rand"
+	"time"
+)
+
+const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+
+// RandomString generates a random strings with a defined length.
+func RandomString(n int) string {
```
ppapapetrou76: At some point this needs to be moved to `cloud-sdk-go` . Maybe we should add a `TODO` 
Also any reason why this is not in the `pkg` directory. It's not something dependent on the cmd layer. 
It's not a blocker for sure - just to keep things tidy

marclop: you're right, I just moved it under `pkg`

PR #55, opened 2019-11-19
deployment: Add update command

marclop:
## Description

Adds the `deployment update` command which mainly relies on a JSON file
defnition as the payload to which the specified cluster ID is going to
be updated.

By default it sets prune_orphans=false in the API payload, which causes
the JSON formatted deployment update payload to be treated as a partial
update, meaning, any resources not explicitly specified, will be ignored
and not shut down. To override this behavior toggle `--prune-orphans`
flag.

## How Has This Been Tested?

(See the markdown help file with the examples.)

## Types of Changes


cmd/deployment/update.go
```diff
@@ -0,0 +1,179 @@
...
+		var r models.DeploymentUpdateRequest
+		if err := cmdutil.DecodeFile(filename, &r); err != nil {
+			return err
+		}
+
+		pruneOrphans, _ := cmd.Flags().GetBool("prune-orphans")
+		r.PruneOrphans = &pruneOrphans
+
+		force, _ := cmd.Flags().GetBool("force")
+		var msg = `setting --prune-orphans to "true" will cause any resources not specified in the update request to be removed from the deployment, do you want to continue? [y/n]: `
```
ppapapetrou76: ❤️

cmd/deployment/update.go
```diff
@@ -0,0 +1,179 @@
...
+
+		return ecctl.Get().Formatter.Format("", res)
+	},
+}
+
+func init() {
+	Command.AddCommand(updateCmd)
+	updateCmd.Flags().Bool("prune-orphans", false, "When set to true, it will remove any resources not specified in the update request, treating the json file contents as the authoritative deployment definition")
+	updateCmd.Flags().Bool("skip-snapshot", false, "Skips taking an Elasticsearch snapshot prior to shutting down the deployment")
+	updateCmd.Flags().Bool("hide-pruned-orphans", false, "Hides orphaned resources that were shut down (only relevant if --skip-orphan-prunning=true)")
```
ppapapetrou76: I don't think we have a `--skip-orphan-prunning` argument, do we? maybe `prune-orphans` instead?

pkg/deployment/update.go
```diff
@@ -0,0 +1,91 @@
...
+	*api.API
+
+	DeploymentID string
+	Request      *models.DeploymentUpdateRequest
+
+	// Optional values
+	SkipSnapshot      bool
+	HidePrunedOrphans bool
+
+	Region string
```
ppapapetrou76: Should we check for Region emptiness?

marclop: Not really since it's an optional arg which gets ignored if it's not. I'll add a comment

cmd/deployment/update.go
marclop: I'll update this flag's description

PR #57, opened 2019-11-21
deployment: Fix show resource type command

### LINKED ISSUE #56: deployment show doesn't display the deployment resources
## Expected Behavior

When I issue `ecctl deployment show <id> --type=<type>` I expect to see the <type> resources tied to my deployment.

## Current Behavior

Currently `main-<type>` is hardcoded as the `ref_id` which is causing this command to not work. This is consistent across the board with all resource types.

```console
$ ecctl deployment show --type elasticsearch d379a4cc6dd74dada234dcfe447c7b48
{
  "errors": [
    {
      "code": "deployments.deployment_resource_not_found",
      "fields": null,
      "message": "Deployment [d379a4cc6dd74dada234dcfe447c7b48] does not have a Resource defined by Kind [elasticsearch] with RefId [main-elasticsearch]"
    }
  ]
}
```

## Possible Solution

I believe that we can provide `ref_id` auto-resolution out of the door via obtaining the whole deployment beforehand and extracting the resource type and its `ref_id`. In case there's multiple `ref_id` resources, a choice can be provided to the user, or all the resources with all `ref_id` can be printed.

Additionally a `--ref-id` flag can be provided so that the user can control this behavior.

## Steps to Reproduce

1. Simply issue `ecctl deployment show --type elasticsearch <deployment id>`

## Context

 As the code is laid out from elastic/ecctl#14, this will only work if the resource `"ref_id"` is `main-<type>` which is not always the case since the user has the ability to change the `ref_id` from deployment to deployment and even the default resources created by the ECE installer

## Your Environment

* Version used: `ecctl 1.0.0-rc1 (build 363e6cb39c5a44b2413d6c2a630881c60bc58839)`

marclop:
## Description

Fixes the the `deployment show --type=<type>` by removing the harcoded
API calls with the RefId set to `main-<type>`. Additionally solves the
problem of an unspecified (by default) RefId to be autodiscovered with
an API call which obtains the RefId of the resource type when Type is
specified but --ref-id is not.

Additionally adds a `--ref-id` flag which allows the users to override
the default "auto-discovery" behavior and obtain a specific resource if
the RefId is known by the user.

Furthermore, moves all the switch cases to obtain different kind of
resources from the deployment to the more appropriate applicaiton logic
package rather than the `cmd` presentation layer.

Adds an example of this behavior to the help/docs.

## Related Issues

Resolves #56.

## Types of Changes

### DIFF
--- a/cmd/deployment/show.go
+++ b/cmd/deployment/show.go
@@ -30,7 +30,10 @@ import (
 
 const showExample = `
 * Shows kibana resource information from a given deployment:
-  ecctl deployment show <deployment-id> --type kibana`
+  ecctl deployment show <deployment-id> --type kibana
+
+* Shows apm resource information from a given deployment with a specified ref-id.
+  ecctl deployment show <deployment-id> --type apm --ref-id apm-server`
 
 var acceptedTypes = []string{"apm", "appsearch", "elasticsearch", "kibana"}
 
@@ -52,9 +55,11 @@ var showCmd = &cobra.Command{
 		plans, _ := cmd.Flags().GetBool("plans")
 		showPlans := planLogs || planDefaults || plans
 
+		refID, _ := cmd.Flags().GetString("ref-id")
 		getParams := deployment.GetParams{
 			API:          ecctl.Get().API,
 			DeploymentID: args[0],
+			RefID:        refID,
 			QueryParams: deputil.QueryParams{
 				ShowPlans:        showPlans,
 				ShowPlanLogs:     planLogs,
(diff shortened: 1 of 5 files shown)


cmd/deployment/show.go
```diff
@@ -89,6 +84,7 @@ var showCmd = &cobra.Command{
 func init() {
 	Command.AddCommand(showCmd)
 	showCmd.Flags().String("type", "", "Optional deployment type to show resource information (elasticsearch, kibana, apm, or appsearch)")
+	showCmd.Flags().String("ref-id", "", "Optional deployment type RefId, if not set, the RefId will be auto-discovered through an API call")
```
ppapapetrou76: I guess `through an API call` is not needed here

pkg/deployment/get_resource.go
```diff
@@ -0,0 +1,112 @@
...
+
+// GetResourceParams is consumed by GetResource.
+type GetResourceParams struct {
+	GetParams
+
+	Type string
+}
+
+// Validate ensures that the parameters are usable by the consuming function.
+func (params GetResourceParams) Validate() error {
```
ppapapetrou76: Do we really need this method here?

marclop: so right

PR #58, opened 2019-11-22
elasticsearch: Add keystore management commands

marclop:
## Description

Adds two new commands, which allow an Elasticsearch cluster's keystore
to be updated and to be shown:

* ecctl deployment elasticsearch keystore set <cluster id> -f def.json
* ecctl deployment elasticsearch keystore show <cluster id>

The specified keystore definition is treated as a partial and the values
in it are used to either create / update the keystore values. Existing
keystore values will not be affected if missing from the request.

## Motivation and Context

Missing elasticsearch keystore management commands.

## Types of Changes

### DIFF
--- a/cmd/deployment/elasticsearch/keystore/get.go
+++ b/cmd/deployment/elasticsearch/keystore/get.go
@@ -1,0 +1,60 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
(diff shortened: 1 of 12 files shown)


cmd/deployment/elasticsearch/keystore/get.go
```diff
@@ -0,0 +1,60 @@
...
+    },
+    "s3.client.foobar.secret_key": {
+      "as_file": false
+    }
+  }
+}`
+
+var getCmd = &cobra.Command{
+	Use:     `show <cluster id>`,
+	Short:   "Shows the current Elasticsearch keystore settings",
```
ppapapetrou76: I think `current` is a bit ambiguous here? maybe `Shows an Elasticsearch cluster's keystore settings`

cmd/deployment/elasticsearch/keystore/set.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/elastic/cloud-sdk-go/pkg/models"
+	"github.com/spf13/cobra"
+
+	cmdutil "github.com/elastic/ecctl/cmd/util"
+	"github.com/elastic/ecctl/pkg/deployment/elasticsearch"
+	"github.com/elastic/ecctl/pkg/ecctl"
+)
+
+const setKeystoreLong = `Manages the keystore settings of an Elasticsearch cluster.
+Note that the underlying API call uses the PATCH method, meaning that each operation is add/modify only.`
```
ppapapetrou76: not sure if we want to provide that level of details to the public. I mean we should clearly mention that this is `add/modify` and not replace but `API` and `PATCH` terms don't seem good as part of users help, no?

marclop:
@ppapapetrou76 hopefully addressed your comments in 69cd902

PR #68, opened 2019-11-29
Remove brew pin-tap step

zanbel:
Removed `brew tap-pin elastic/ecctl` step from the macOS install guide as it results in as it's not required and causes an error.


elasticcla:
Hi @zanbel, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ecctl/pull/68.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

zanbel:
@elasticcla  I've updated the emails list to include roy@elastic.co.

PR #72, opened 2019-12-04
cmd: Add deployment plan cancel

marclop:
## Description

Adds the deployment plan cancel command which allows users to cancel
pending plans for specific resources.
The syntax is as follows:
`ecctl deployment plan cancel <deployment id> --type <type> --ref-id <id>`

## Motivation and Context

Implement more of the Deployment resource APIs.

## How Has This Been Tested?

```console
$ ecctl deployment plan cancel fc2fe19a8c4f43d385004febe4e63900 --type kibana --ref-id kibana --force
{
  "errors": [
    {
      "code": "deployments.resource_does_not_have_a_pending_plan",
      "fields": null,
      "message": "The Resource does not have a pending plan: DeploymentId [fc2fe19a8c4f43d385004febe4e63900], Kind [kibana], ResourceId: [8d37cff942fb482ba7a3cac10eea943a]"
    }
  ]
}
```

## Types of Changes

### DIFF
--- a/cmd/deployment/plan/command.go
+++ b/cmd/deployment/plan/command.go
@@ -30,6 +30,3 @@ var Command = &cobra.Command{
 		cmd.Help()
 	},
 }
-
-func init() {
-}
--- a/cmd/deployment/plan/command.go
+++ b/cmd/deployment/plan/command.go
@@ -15,22 +15,16 @@
 // specific language governing permissions and limitations
 // under the License.
 
-package cmddeployment
+package cmddeploymentplan
 
 import (
 	"github.com/spf13/cobra"
-
-	cmdapm "github.com/elastic/ecctl/cmd/deployment/apm"
-	cmdelasticsearch "github.com/elastic/ecctl/cmd/deployment/elasticsearch"
-	cmdkibana "github.com/elastic/ecctl/cmd/deployment/kibana"
-	cmddeploymentnote "github.com/elastic/ecctl/cmd/deployment/note"
-	cmddeploymentplan "github.com/elastic/ecctl/cmd/deployment/plan"
 )
 
 // Command is the deployment subcommand
 var Command = &cobra.Command{
-	Use:     "deployment",
-	Short:   "Manages deployments",
+	Use:     "plan",
+	Short:   "Manages deployment plans",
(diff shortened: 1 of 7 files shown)


cmd/deployment/plan/command.go
```diff
@@ -0,0 +1,35 @@
...
+var Command = &cobra.Command{
+	Use:     "plan",
+	Short:   "Manages deployment plans",
+	PreRunE: cobra.MaximumNArgs(0),
+	Run: func(cmd *cobra.Command, args []string) {
+		cmd.Help()
+	},
+}
+
+func init() {
```
infraredgirl: Is the empty `init()` necessary?

pkg/deployment/depresource/cancel_plan.go
```diff
@@ -0,0 +1,80 @@
...
+
+	if params.API == nil {
+		merr = multierror.Append(merr, util.ErrAPIReq)
+	}
+
+	if len(params.DeploymentID) != 32 {
+		merr = multierror.Append(merr, deputil.NewInvalidDeploymentIDError(params.DeploymentID))
+	}
+
+	if params.Type == "" {
```
ppapapetrou76: I'm wondering if it makes any sense to add some checks here about the accepted types. I know we are using this in a few places in the code so we could create a common check. Ignore me if it doesn't worth the time

pkg/deployment/depresource/cancel_plan.go
```diff
@@ -0,0 +1,80 @@
...
+	*api.API
+
+	DeploymentID string
+	Type         string
+	RefID        string
+	ForceDelete  bool
+}
+
+// Validate ensures the parameters are usable by the consuming function.
+func (params CancelPlanParams) Validate() error {
```
ppapapetrou76: We are missing test for this method. I think we should explicitly test this function as we already do in other places rather than indirectly test it

pkg/deployment/depresource/cancel_plan.go
marclop: We discussed it out of band and agreed that we should't overload this client logic and leave the API to validate the type sent as a queryparam

PR #79, opened 2019-12-10
cmd: Migrate apm create to deployments API

### LINKED ISSUE #73: Move APM Create to Deployments API
## Overview

`ecctl deployment apm create` allows both the creation of a simple Kibana deployment and of a complex _sliders_-able cluster with a more complex topology. There's a number of problems with the APIs and the way that command operates:

1. Uses `POST /v1/clusters/apm` to create it, which we should move away from since it'll soon be deprecated.
2. Creating a apm with `ecctl deployment apm create` only supports a file definition and as a simpler way to create a Kibana deployment `ecctl deployment apm enable` does so with a default memory and topology which can't be changed.

marclop:
## Description

This change mainly moves the apm create action to use the deployment
API instead of creating the cluster through the cluster APIs.
It makes use of the deployment update API to create the apm resource
within a deployment which should already exist and have an elasticsearch
resource.

It only provides simple flags and a flat topology with a specified size
for the node memory and the number of zones the deployment resource will
span.

It moves the common arguments to a structure named NewStateless
rather than having individual type-specific parameter structures.

Additionally, adds a `cmdutil.Track` function which handles the formatter
and tracking mix of things.

## Related Issues

Resolves #73

## How Has This Been Tested?

(See markdown examples)

## Types of Changes


pkg/deployment/depresource/apm_test.go
```diff
@@ -0,0 +1,217 @@
...
+							}},
+						},
+					})),
+				),
+				Region: "ece-region",
+			}},
+			err: errors.New("unable to obtain deployment template ID from existing deployment ID, please specify a one"),
+		},
+		{
+			name: "obtains the deployment info but fails getting the template ID info",
```
ppapapetrou76: I think this description here is the same with the previous test

pkg/deployment/depresource/kibana_test.go
```diff
@@ -129,7 +129,7 @@ func TestNewKibana(t *testing.T) {
 		},
 		{
 			name: "obtains the deployment info but fails getting the template ID info",
```
ppapapetrou76: I think this description here is the same with the previous test

pkg/deployment/depresource/new_stateless.go
```diff
@@ -0,0 +1,106 @@
...
+	// Required: DeploymentID.
+	DeploymentID string
+
+	// Required region name.
+	Region string
+
+	// Optional ElasticsearchRefID.
+	ElasticsearchRefID string
+
+	// Optional name. If not specified it defaults to the autogeneratd ID.
```
ppapapetrou76:
```suggestion
	// Optional name. If not specified it defaults to the autogenerated ID.
```

pkg/deployment/depresource/new_stateless.go
```diff
@@ -0,0 +1,106 @@
...
+	Name string
+
+	// Optional RefID for the for the stateless deployment resource.
```
ppapapetrou76:
```suggestion
	// Optional RefID for the stateless deployment resource.
```

cmd/deployment/apm/create.go
```diff
@@ -18,59 +18,123 @@
...
-		})
+		}
+
+		res, err := deployment.Update(updateParams)
 		if err != nil {
 			return err
 		}
-		return ecctl.Get().Formatter.Format(filepath.Join("apm", "create"), res)
+
+		if err := ecctl.Get().Formatter.Format("", res); err != nil {
```
ppapapetrou76: I have the same concern here about duplicating the tracking code in several places

cmd/deployment/apm/create.go
```diff
@@ -18,59 +18,123 @@
...
+	"github.com/elastic/ecctl/pkg/deployment/depresource"
 	"github.com/elastic/ecctl/pkg/ecctl"
-	"github.com/elastic/ecctl/pkg/util"
 )
 
 var createApmCmd = &cobra.Command{
-	Use:     "create -f <file definition> --id <deployment id>",
-	Short:   "Creates an APM instance in the selected deployment",
+	Use:     "create -f <definition.json>",
+	Short:   "Creates a Apm instance",
```
karencfv:
```suggestion
	Short:   "Creates an APM instance",
```

cmd/deployment/apm/create.go
```diff
@@ -18,59 +18,123 @@
...
 
 func init() {
-	createApmCmd.Flags().StringP("file", "f", "", "Deployment definition")
-	createApmCmd.Flags().String("id", "", "Overrides the deployment ID on which it's tied")
-	createApmCmd.Flags().Bool("track", false, cmdutil.TrackFlagMessage)
+	Command.AddCommand(createApmCmd)
+	createApmCmd.Flags().BoolP("track", "t", false, cmdutil.TrackFlagMessage)
+	createApmCmd.Flags().StringP("file", "f", "", "ApmPayload file definition. See help for more information")
+	createApmCmd.Flags().String("deployment-template", "", "Optional deployment template ID, automatically obtained from the current deployment")
+	createApmCmd.Flags().String("version", "", "Version to use, if not specified, the deployment's stack version will be used")
```
karencfv:
```suggestion
	createApmCmd.Flags().String("version", "", "Optional version to use. If not specified, it will default to the deployment's stack version")
```

cmd/deployment/apm/create_help.go
```diff
@@ -0,0 +1,77 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package cmdapm
+
+const (
+	apmCreateLong = `Creates an APM deployment, limitting the creation scope to APM resources.
+There's a few ways to create an APM deployment, sane default values are provided, making
```
karencfv:
```suggestion
There are a few ways to create an APM deployment. Sane default values are provided, making
```

cmd/deployment/apm/create_help.go
```diff
@@ -0,0 +1,77 @@
...
+the command work out of the box even when no parameters are set. When version is not specified,
+the matching elasticsearch deployment version will be used. These are the available options:
+
+  * Simplified flags: --zones <zone count> --size <node memory in MB>
+  * File definition: --file=<file path> (shorthand: -f). The definition can be found in:
+    https://www.elastic.co/guide/en/cloud-enterprise/current/definitions.html#ApmPayload
+
+As an option "--generate-payload" can be used in order to obtain the generated ApmPayload
```
karencfv:
```suggestion
As an option, "--generate-payload" can be used in order to obtain the generated ApmPayload
```

PR #80, opened 2019-12-10
cmd: Add --track flag to deployment commands

### LINKED ISSUE #78: Add --track flag deployment commands which incur plan changes
## Overview

Currently, deployment subcommands have no `--track` flag.

## Possible Implementation

Add a `--track` flag so that the resource creation / change can be tracked

marclop:
## Description

Adds the --track flag and resource tracking after a plan change to the
deployment subcommands which incur plan changes: `create`, `update` and
`shutdown`.

## Related Issues

Resolves #78 

## How Has This Been Tested?

Manually

## Types of Changes

### DIFF
--- a/cmd/deployment/create.go
+++ b/cmd/deployment/create.go
@@ -170,21 +170,15 @@ var createCmd = &cobra.Command{
 		}
 
 		var track, _ = cmd.Flags().GetBool("track")
-		if err := ecctl.Get().Formatter.Format("", res); err != nil {
-			if !track {
-				return err
-			}
-			fmt.Fprintln(ecctl.Get().Config.OutputDevice, err)
-		}
-
-		if !track {
-			return nil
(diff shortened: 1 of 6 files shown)


cmd/deployment/create.go
```diff
@@ -167,12 +169,29 @@ var createCmd = &cobra.Command{
 			return err
 		}
 
-		return ecctl.Get().Formatter.Format("", res)
+		var track, _ = cmd.Flags().GetBool("track")
```
ppapapetrou76: I think this one deserves a common function `TrackOrReturn`. The code is almost identical and probably we will have more code like this in the future. I know we already have existing code like this so it makes more sense to keep it in a single place

marclop:
I'll rebase after #79 is merged to use the common function introduced there.

PR #86, opened 2019-12-12
cmd: Add deployment resource shutdown command

marclop:
## Description

Adds a `ecctl deployment resource shutdown` command which allows
specific deployment resources to be targetted for shutdown. It requires
two flags: `--ref-id` and `--type` which the user needs to specify as
per the shown help.

Additionally, two optional flags `--hide` and `--skip-snapshot` are
available to toggle those options

Furthermore, it extracts the common fields `API`, `DeploymentID`, `Type`
and `RefID`, adding RefID auto-discovery when not specified.

## How Has This Been Tested?

Manually.

## Types of Changes

### DIFF
--- a/pkg/deployment/depresource/shutdown.go
+++ b/pkg/deployment/depresource/shutdown.go
@@ -18,23 +18,15 @@
 package depresource
 
 import (
-	"errors"
-
-	"github.com/elastic/cloud-sdk-go/pkg/api"
 	"github.com/elastic/cloud-sdk-go/pkg/client/deployments"
-	"github.com/hashicorp/go-multierror"
 
-	"github.com/elastic/ecctl/pkg/deployment/deputil"
+	"github.com/elastic/ecctl/pkg/deployment"
 	"github.com/elastic/ecctl/pkg/util"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/deployment/depresource/shutdown.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/elastic/ecctl/pkg/deployment/deputil"
+	"github.com/elastic/ecctl/pkg/util"
+)
+
+// ShutdownParams is consumed by Shutdown.
+type ShutdownParams struct {
+	*api.API
+
+	DeploymentID string
```
ppapapetrou76: I'm seeing that most of the resources commands have those four fields the same  (`API`,`DeploymentID`,`RefID`, `Type` ) and we perform the same validations 

Would it make sense to extract those for into a `CommonXTZParams` struct with the its own validation and embed it to other structs?

This will save us a lot of code and increase code re-use and avoidance of future errors ( for instance add one more check or field )

marclop:
@ppapapetrou76 This is ready for review again

PR #87, opened 2019-12-12
cmd: Add deployment resource restore command

marclop:
## Description

Adds an `ecctl deployment resource restore` command which restores a
previously shut down individual deployment resource. If ref-id not specified
it's auto-discovered via an API call.

Required flags: `--type`.
Optional flag: `--restore-snapshot` and `--ref-id` 

## How Has This Been Tested?

Manually

## Types of Changes

### DIFF
--- a/cmd/deployment/resource/restore.go
+++ b/cmd/deployment/resource/restore.go
@@ -24,6 +24,7 @@ import (
 	"github.com/spf13/cobra"
 
 	cmdutil "github.com/elastic/ecctl/cmd/util"
+	"github.com/elastic/ecctl/pkg/deployment"
 	"github.com/elastic/ecctl/pkg/deployment/depresource"
 	"github.com/elastic/ecctl/pkg/ecctl"
 )
@@ -53,10 +54,12 @@ var restoreCmd = &cobra.Command{
 		}
 
 		return depresource.Restore(depresource.RestoreParams{
-			API:             ecctl.Get().API,
-			DeploymentID:    args[0],
-			Type:            resType,
-			RefID:           refID,
+			ResourceParams: deployment.ResourceParams{
+				API:          ecctl.Get().API,
+				DeploymentID: args[0],
+				Type:         resType,
+				RefID:        refID,
+			},
 			RestoreSnapshot: restoreSnapshot,
 		})
 	},
@@ -66,7 +69,6 @@ func init() {
 	Command.AddCommand(restoreCmd)
 	restoreCmd.Flags().String("type", "", "Required deployment type to restore (elasticsearch, kibana, apm, or appsearch)")
 	restoreCmd.MarkFlagRequired("type")
-	restoreCmd.Flags().String("ref-id", "", "Required deployment RefId")
-	restoreCmd.MarkFlagRequired("ref-id")
+	restoreCmd.Flags().String("ref-id", "", "Optional deployment RefId, auto-discovered if not specified")
 	restoreCmd.Flags().Bool("restore-snapshot", false, "Optional flag to toggle restoring a snapshot for an Elasticsearch resource. It has no effect for other resources")
 }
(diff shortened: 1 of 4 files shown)


cmd/deployment/resource/restore.go
```diff
@@ -0,0 +1,64 @@
...
+			Type:            resType,
+			RefID:           refID,
+			RestoreSnapshot: restoreSnapshot,
+		})
+	},
+}
+
+func init() {
+	Command.AddCommand(restoreCmd)
+	restoreCmd.Flags().String("type", "", "Required stateless deployment type to restore (elasticsearch, kibana, apm, or appsearch)")
```
ppapapetrou76:
```suggestion
	restoreCmd.Flags().String("type", "", "Required deployment type to restore (elasticsearch, kibana, apm, or appsearch)")
```

pkg/deployment/depresource/restore.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Restore upgrades a stateless deployment resource like APM, Kibana
+// and AppSearch.
+func Restore(params RestoreParams) error {
+	if err := params.Validate(); err != nil {
+		return err
+	}
+
+	return util.ReturnErrOnly(params.V1API.Deployments.RestoreDeploymentResource(
```
ppapapetrou76: What is the behavior of the API when the type is not `elasticsearch` and we turn on the `--restore-snapshost` flag? 
Do we get a meaningful error?
Do we want to show a message to the user (on the cmd layer) that the flag will be ignored?

cmd/deployment/resource/restore.go
```diff
@@ -0,0 +1,64 @@
...
+	Short:   "Restores a previously shut down deployment resource",
+	PreRunE: cmdutil.MinimumNArgsAndUUID(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		resType, _ := cmd.Flags().GetString("type")
+		refID, _ := cmd.Flags().GetString("ref-id")
+		restoreSnapshot, _ := cmd.Flags().GetBool("restore-snapshot")
+
+		force, _ := cmd.Flags().GetBool("force")
+		var dataLoss = resType == "elasticsearch" && !restoreSnapshot
+		var msg = "This action restore an Elasticsearch resource without its snapshot which might incur data loss, do you want to continue? [y/n]: "
```
karencfv:
```suggestion
		var msg = "This action restores an Elasticsearch resource without its snapshot, which might incur data loss. Do you want to continue? [y/n]: "
```

pkg/deployment/depresource/restore.go
marclop: I tried and there's no API error and I would expect it so since the API accepts that restore_snapshot parameter by default:
```
 deployments.NewRestoreDeploymentResourceParams().
 	WithRestoreSnapshot(&params.RestoreSnapshot).
```

I could still add a prompt stating that the action won't have any effect, tho

marclop: 38f80f3 should address it

cmd/deployment/resource/restore.go
```diff
@@ -0,0 +1,72 @@
...
+	},
+}
+
+func init() {
+	Command.AddCommand(restoreCmd)
+	restoreCmd.Flags().String("type", "", "Required deployment type to restore (elasticsearch, kibana, apm, or appsearch)")
+	restoreCmd.MarkFlagRequired("type")
+	restoreCmd.Flags().String("ref-id", "", "Required deployment RefId")
+	restoreCmd.MarkFlagRequired("ref-id")
+	restoreCmd.Flags().Bool("restore-snapshot", false, "Optional flag to toggle restoring a snapshot for an Elasticsearch resource. It has no eeffect for other resources")
```
ppapapetrou76:
```suggestion
	restoreCmd.Flags().Bool("restore-snapshot", false, "Optional flag to toggle restoring a snapshot for an Elasticsearch resource. It has no effect for other resources")
```

marclop:
Added ref-id autodiscovery in b93742c

PR #88, opened 2019-12-12
cmd: Add deployment resource delete command

marclop:
## Description

Adds an `ecctl deployment resource delete` command which deletes a
previously shut down deployment resource. If ref-id not specified
it's auto-discovered via an API call.

Required flags: `--type`
Optional flags: `--ref-id`

## How Has This Been Tested?

Manually

## Types of Changes

### DIFF
--- a/cmd/deployment/resource/delete.go
+++ b/cmd/deployment/resource/delete.go
@@ -1,0 +1,52 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package cmddeploymentresource
+
+import (
+	"github.com/spf13/cobra"
+
+	cmdutil "github.com/elastic/ecctl/cmd/util"
+	"github.com/elastic/ecctl/pkg/deployment/depresource"
+	"github.com/elastic/ecctl/pkg/ecctl"
+)
+
+// deleteCmd is the deployment subcommand
+var deleteCmd = &cobra.Command{
+	Use:     "delete <deployment id> --type <type> --ref-id <ref-id>",
+	Short:   "Deletes a previously shut down deployment resource",
+	PreRunE: cmdutil.MinimumNArgsAndUUID(1),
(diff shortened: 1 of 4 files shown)


pkg/deployment/depresource/delete_stateless.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if params.RefID == "" {
+		merr = multierror.Append(merr, errors.New("deployment resource ref id cannot be empty"))
+	}
+
+	if params.Type == "" {
+		merr = multierror.Append(merr, errors.New("deployment resource type cannot be empty"))
+	}
+
+	if params.Type == "elasticsearch" {
```
ppapapetrou76: I'm confused again 😄 are we or are we not doing any `type` validation at last?

cmd/deployment/resource/delete.go
```diff
@@ -0,0 +1,52 @@
...
+		})
+	},
+}
+
+func init() {
+	Command.AddCommand(deleteCmd)
+	deleteCmd.Flags().String("type", "", "Required stateless deployment type to upgrade (kibana, apm, or appsearch)")
+	deleteCmd.MarkFlagRequired("type")
+	deleteCmd.Flags().String("ref-id", "", "Required deployment RefId")
+	deleteCmd.MarkFlagRequired("ref-id")
```
karencfv: I believe `ref_id` discovery should implemented in case this flag is not defined, instead of making the flag required.

marclop: Well, I'm not so positive in this case since it's a destructive action. The reason why I didn't implement it was precisely because of that. If we feel we still want to go forward with `ref_id` auto-discovery, then I'll add it

pkg/deployment/depresource/delete_stateless.go
marclop: The reason here is that this command is "stateless" so Elasticsearch will never work as an argument. The only type validation I added. I can remove it if it seems like we're locking down into it

cmd/deployment/resource/delete.go
marclop: I thought that maybe what we want to do is return an error when there's more than 1 resource `ref_id` which is unlikely but could happen.

Say that the _auto_ ref_id discovery yields more than 1 ref_id for a resource type then we could just return an error saying that 2 ref_ids have been found for resource type, and ask the user to chose one and re-run the command.

karencfv: The reason I mentioned this is because of the possibility that a user might not be completely aware of what `ref_id` is, if they normally create their deployments with defaults. Maybe asking to confirm action like we'll be doing for `stop` and `start-maintenance` is a good idea?

> I thought that maybe what we want to do is return an error when there's more than 1 resource ref_id which is unlikely but could happen.

I think we should definitely do this as well. It is very likely that this will happen given that the UI defaults to adding a `main-` prefix and `ecctl` does not. So someone using both UI and ecctl will have mixed defaults (e.g. `ref_id: main-apm` for the first instances if they used the UI to create the first deployment and `ref_id: apm` if they used ecctl to add more instances later on)

marclop: I addressed your concerns in 0601d61 after rebasing with the newly added `deployment.ResourceParams`

Issue #83, opened 2019-12-11
Add CRUD commands for deployment resources


marclop:
## Description
Now that the swagger spec in cloud-sdk-go has been [updated](https://github.com/elastic/cloud-sdk-go/pull/9) to the latest changes in 2.4 we need to implement the new CRUD endpoints for the deployments api that relate to resources with the following IDs:

- [x] cancel-deployment-resource-pending-plan (@marclop, elastic/ecctl#72)
- [x] delete-deployment-stateless-resource (@marclop, #88)
- [x] upgrade-deployment-stateless-resource (@marclop, https://github.com/elastic/ecctl/pull/76)
- [x] get-deployment-apm-resource-info (@karencfv, https://github.com/elastic/ecctl/pull/14)
- [x] get-deployment-es-resource-info (@karencfv, https://github.com/elastic/ecctl/pull/14)
- [x] get-deployment-kib-resource-info (@karencfv, https://github.com/elastic/ecctl/pull/14)
- [x] set-deployment-resource-raw-metadata (@karencfv )
- [x] restore-deployment-resource (@marclop, #87)
- [x] shutdown-deployment-es-resource (@marclop, #86)
- [x] shutdown-deployment-stateless-resource (@marclop, #86)
- [x] start-deployment-resource-instances (@karencfv https://github.com/elastic/ecctl/pull/89)
- [x] start-deployment-resource-instances-all (@karencfv https://github.com/elastic/ecctl/pull/89)
- [x] start-deployment-resource-instances-all-maintenance-mode (@karencfv #89 )
- [x] start-deployment-resource-maintenance-mode (@karencfv #89 )
- [x] stop-deployment-resource-instances (@karencfv, #81)
- [x] stop-deployment-resource-instances-all (@karencfv, #81)
- [x] stop-deployment-resource-instances-all-maintenance-mode (@karencfv, #81)
- [x] stop-deployment-resource-maintenance-mode (@karencfv, #81)

marclop:
I'm currently working on the implementation of a command which calls `shutdown-deployment-es-resource` or `shutdown-deployment-stateless-resource` depending on which resource type has been set for shutdown.

karencfv:
working on ` start-deployment-resource-instances-` related commands :)

karencfv:
working on `set-deployment-resource-raw-metadata`

marclop:
Work has been complete, thanks @karencfv !!

PR #89, opened 2019-12-12
cmd: add deployment resource start/start-maintenance commands

karencfv:
## Description
Adds the following command:

- `ecctl deployment resource start` which starts a previously stopped deployment resource.
- `ecctl deployment resource start-maintenance` which starts maintenance mode on a deployment resource.

## Related Issue
https://github.com/elastic/ecctl/issues/83

## Motivation and Context
Implementing more of the deployment/resource APIs.

## How Has This Been Tested?
Manually against an ece environment and by running unit test suite.

## Types of Changes

- [ ] Documentation

## Readiness Checklist


cmd/deployment/resource/start-maintenance.go
```diff
@@ -0,0 +1,74 @@
...
+var startMaintCmd = &cobra.Command{
+	Use:     "start-maintenance <deployment id> --type <type> [--all|--i <instance-id>,<instance-id>]",
+	Short:   "Starts maintenance mode on a deployment resource",
+	PreRunE: cmdutil.MinimumNArgsAndUUID(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		resType, _ := cmd.Flags().GetString("type")
+		refID, _ := cmd.Flags().GetString("ref-id")
+		instanceID, _ := cmd.Flags().GetStringSlice("instance-id")
+		ignoreMissing, _ := cmd.Flags().GetBool("ignore-missing")
+		all, _ := cmd.Flags().GetBool("all")
```
ppapapetrou76: should we add the same prompt as @marclop suggested in the other PR? Setting all instances to maintenance mode will cause downtime.

pkg/deployment/depresource/start.go
```diff
@@ -0,0 +1,164 @@
...
+	*api.API
+	DeploymentID string
+	Type         string
+	RefID        string
+	All          bool
+}
+
+// Validate ensures the parameters are usable by Start.
+func (params StartParams) Validate() error {
+	var merr = new(multierror.Error)
```
ppapapetrou76: I think we have a common structure right @marclop to avoid duplication?

pkg/deployment/depresource/start.go
```diff
@@ -0,0 +1,164 @@
...
+	StartParams
+	IgnoreMissing *bool
+	InstanceIDs   []string
+}
+
+// Validate ensures the parameters are usable by StartInstances.
+func (params StartInstancesParams) Validate() error {
+	var merr = new(multierror.Error)
+
+	if len(params.InstanceIDs) == 0 {
```
ppapapetrou76: Is this valid if `all` flag is used?

karencfv: That code isn't touched when the `--all` flag is set https://github.com/elastic/ecctl/pull/89/files#diff-e9b68f402a2f2dd49799010fb1b725f1R78-R79 :)

pkg/deployment/depresource/start.go
karencfv: Do you mean `depresource.NewStateless{}` ? I don't think that struct works very well here, It has a bunch of fields that aren't relevant, and also the `Validate()` and `fillDefaults()` methods are different.

I can make a common struct for these [commands](https://github.com/elastic/ecctl/pull/89/files#diff-037fd59d86ef9ef3885b80d1b6126216R33) and [`stop`/`stop-maintenance`](https://github.com/elastic/ecctl/pull/81/files#diff-ccddd9d334f1b421e58829b51a7593ccR33) called `StopStartParams` if you want.

ppapapetrou76: No , no , I can't find a discussion I had with Marc last week on a GH pr about the same. @marclop can u help pls? :)

marclop: We discussed out of band that going forward probably a common structure with: `API`, `DeploymentID`, `RefID`, `Type` can be defined and used with all the validations in common. Including the fillDefaults ref-id auto-discovery.

ppapapetrou76: Ok so there will be another PR introducing the common struct with validation + filldefaults. 
Great. we are good to go then

Issue #91, opened 2019-12-16
RefID is not auto-dicovered with `ecctl deployment resource upgrade`


karencfv:
## Readiness Checklist

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
As per the command's flag description `RefID` should be auto-discovered when the flag is not set.

```cosnole
--ref-id string   Optional deployment RefId, if not set, the RefId will be auto-discovered
```

## Current Behavior
When the command is run, it tells me it is required to provide a `ref-id`

```console
$ go run main.go --config ce-aws deployment resource upgrade fba15d094a8a4a7983c32b7d0dc58596 --type apm
required flag(s) "ref-id" not set
exit status 255
```

## Possible Solution
Have the command auto-discover RefID

## Steps to Reproduce

1.  run `go run main.go deployment resource upgrade <cluster-id> --type kibana`

PR #92, opened 2019-12-16
cmd: Add ref-id auto-discovery to resource upgrade

karencfv:
## Description
Adds RefID auto-discovery when the `--ref-id` flag is not set.

## Related Issues
Closes: https://github.com/elastic/ecctl/issues/91

## Motivation and Context
Fixes a bug where this value was not being auto-discovered as per the flag's description.
```console
--ref-id string   Optional deployment RefId, if not set, the RefId will be auto-discovered
```

## How Has This Been Tested?
Manually and with unit tests

```console
$ go run main.go --config ce-aws deployment resource upgrade fba15d094a8a4a7983c32b7d0dc58596 --type apm --verbose
<...>
==================== Start of Request #2 ====================
POST /api/v1/deployments/fba15d094a8a4a7983c32b7d0dc58596/apm/main-apm/_upgrade?validate_only=false HTTP/1.1
<...>
```

## Types of Changes

- [ ] Documentation

## Readiness Checklist


pkg/deployment/depresource/upgrade_stateless_test.go
```diff
@@ -71,14 +140,35 @@ func TestUpgradeStateless(t *testing.T) {
...
+				DeploymentID: util.ValidClusterID,
+				Type:         "elasticsearch",
+			}},
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			got, err := UpgradeStateless(tt.args.params)
-			if !reflect.DeepEqual(err, tt.err) {
+			if tt.err != nil && err.Error() != tt.err.Error() {
```
marclop: I think we should settle on which error assertions we want to follow. It's not a big deal, but we're not asserting the same thing, an error could be returned by the function anyway and we wouldn't catch it unless there's something in the `err` field of the test structure, which won't be if the test structure is not asserting for an error different than `nil`.

karencfv: I prefer using `!reflect.DeepEqual(err, tt.err)` as well :) In this case, unfortunately, when I added the last [test case](https://github.com/elastic/ecctl/pull/92/files#diff-2c76ec0b0bb37f47f4e839bf6f176a2dR148) I started getting the following error, so I had to change the assertion.

```
--- FAIL: TestUpgradeStateless/succeeds_when_RefID_is_not_set (0.00s)
    /Users/karen/go/src/github.com/elastic/ecctl/pkg/deployment/depresource/upgrade_stateless_test.go:172: UpgradeStateless() error = , wantErr <nil>
```

karencfv:
Updated the PR to use `deployment.ResourceParams` :) @ppapapetrou76 @marclop

pkg/deployment/depresource/upgrade_stateless_test.go
marclop: huh that's weird

ppapapetrou76: DeepEquals probably fails because the tests are of different type...

PR #101, opened 2019-12-18
Add auto completion intruction to docs

zanbel:
Added more specific instructions how to enable auto completion in the note following the brew install commands.

### DIFF
--- a/docs/ecctl-getting-started.asciidoc
+++ b/docs/ecctl-getting-started.asciidoc
@@ -46,7 +46,7 @@ Updating Homebrew...
 🍺  /usr/local/Cellar/ecctl/1.0.0: 6 files, 13MB, built in 7 seconds
 ----
 
-NOTE: To get autocompletions working make sure to run `source <(ecctl generate completions)`. To your shell interpreter configuration file by running `echo "source <(ecctl generate completions)" >> ~/.bash_profile` or `echo "source <(ecctl generate completions)" >> ~/.zshrc`.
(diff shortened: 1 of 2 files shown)


elasticcla:
Hi @zanbel, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ecctl/pull/101.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

docs/ecctl-getting-started.asciidoc
```diff
@@ -46,7 +46,7 @@ Updating Homebrew...
 🍺  /usr/local/Cellar/ecctl/1.0.0: 6 files, 13MB, built in 7 seconds
 ----
 
-NOTE: To get autocompletions working, they must be sourced from your shell profile.
+NOTE: To get autocompletions working make sure to run `source <(ecctl generate completions)` and to your shell interpreter configuration file by running `echo "source <(ecctl generate completions)" >> ~/.bash_profile` or `echo "source <(ecctl generate completions)" >> ~/.zshrc`.
```
karencfv:
```suggestion
NOTE: To get autocompletions working make sure to run `source <(ecctl generate completions)`. If you prefer to add to your shell interpreter configuration file run `echo "source <(ecctl generate completions)" >> ~/.bash_profile` or `echo "source <(ecctl generate completions)" >> ~/.zshrc`.
```

nrichers: ++ on the suggested edit from @karencfv. The only other thing to, er, note is that we should `NOTE:` sparingly. We overuse callouts and people start tuning them out. The type of content you have here looks more like something that might into a `Before you begin` section to be consistent with the rest of our ECE and ESS docs.

zanbel:
Thanks for the feedback! Changed according to your suggestion so f you can approve again I'll merge it.

PR #102, opened 2019-12-18
Update ecctl-getting-started.asciidoc

Kushmaro:
We should encourage users to use https endpoints, and ports accordingly.


marclop:
I think we should also add a disclaimer that using https with a self-signed certificate (if it's the case) will require the `--insecure` flag.

Kushmaro:
Sure @marclop, I agree that statement should in the docs generally,
but for this specific init example I believe it should be implicit

PR #103, opened 2019-12-18
cmd: Remove newer version check on ecctl version

marclop:
## Description

Removes the check for a newer version against the current ecctl version
since it requires a GitHub Token. It is not the most useful of features
and not well documented. It would be handy to have the ability to check
for newer versions without any request token or authenticaiton. For the
time being I am removing the check.

## How Has This Been Tested?

`ecctl version` doesn't check for updates anymore.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -8,7 +8,6 @@ third-party software developed by the licenses listed below.
 
 github.com/davecgh/go-spew              0BSD
 github.com/elastic/cloud-sdk-go         Apache-2.0
-github.com/elastic/uptd                 Apache-2.0
 github.com/go-openapi/runtime           Apache-2.0
 github.com/go-openapi/strfmt            Apache-2.0
 github.com/marclop/elasticsearch-cli    Apache-2.0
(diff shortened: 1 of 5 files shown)


infraredgirl:
I understand the reason behind this change, however I still find this feature useful sometimes.

Instead of removing this code, how about putting it behind a `--check-for-new-version` switch that is false by default?

ppapapetrou76:
This is a public / open source tool. IMHO it's a terrible idea to ask people to set an env variable with their GH token ( forcing them to open a GH account if they don't have ) to check for ecctl version.

+1 to completely remove the feature and re-thing another way of checking versions

NOTICE
```diff
@@ -8,20 +8,18 @@ third-party software developed by the licenses listed below.
...
 github.com/pkg/errors                   BSD-2-Clause
 github.com/pmezard/go-difflib           BSD-3-Clause
 golang.org/x/crypto                     BSD-3-Clause
-golang.org/x/oauth2                     BSD-3-Clause
 github.com/asaskevich/govalidator       MIT
 github.com/blang/semver                 MIT
 github.com/chzyer/test                  MIT
 github.com/spf13/viper                  MIT
-github.com/hashicorp/go-multierror      MPL-2.0
+github.com/hashicorp/go-multierror      MPL-2.0-no-copyleft-exception
```
ppapapetrou76: Can we rollback this pls? I know it's auto-generated but....

marclop:
The code removed is trivial and again, I don't think it's the best experience, I take your feedback @karencfv and @infraredgirl, while the functionality provided is useful, the way on which it is, not so much. I have some ideas on how to provide a better experience with this and uptd.

PR #105, opened 2019-12-19
stack: Skip returning an error on packed __MACOSX

marclop:
## Description

Skips returning an error on stackpacks which have a __MACOSX packed into
the .zip file.

## Motivation and Context

ECE stack packs seem to have a __MACOSX packed file which is causing the command to return an error.

```console
$ecctl platform stack upload 7.5.1.zip
1 error occurred:
	* stackpack.invalid_configuration_version: This version cannot be parsed: [__MACOSX] because: Unknown version string: [__MACOSX]
```

## How Has This Been Tested?

Maually

## Types of Changes

### DIFF
--- a/pkg/platform/stack/stack.go
+++ b/pkg/platform/stack/stack.go
@@ -92,6 +92,10 @@ func Upload(params UploadParams) error {
 	var merr = new(multierror.Error)
 	for _, e := range res.Payload.Errors {
 		for _, ee := range e.Errors.Errors {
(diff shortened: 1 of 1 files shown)


infraredgirl:
What a weird bug :smile: 

The code LGTM, I would maybe just add a comment explaining the purpose of the `if` statement, otherwise I think it would be hard for someone reading the code to understand why it's there. The same text that you have under "Motivation and context" I think would suffice.

Issue #99, opened 2019-12-18
Smoke test commands


karencfv:
## Description
Commands should be smoke tested against an ECE environment before the next release.

List to track commands that have been tested:

- [x] auth
    - [x] key create @marclop
    - [x] key delete @marclop
    - [x] key list @marclop
    - [x] key show @marclop
- [x] deployment
    - [x] apm
        - [x] create - Works fine using the simplified flags, but I keep getting an error when trying to use a file definition. Please someone verify that this is an actual error and not something that I might be doing wrong @karencfv, @marclop Fixed in #109
        - [x] delete @karencfv 
        - [x] list @karencfv 
        - [x] plan cancel @karencfv
        - [x] plan history @karencfv
        - [x] plan monitor @karencfv
        - [x] plan reapply @karencfv
        - [x] restart @karencfv
        - [x] resync @karencfv
        - [x] show @karencfv
        - [x] stop @karencfv
        - [x] upgrade @karencfv
    - [x] create @karencfv
    - [x] delete @karencfv
    - [x] elasticsearch
        - [x] console @karencfv 
        - [x] delete @karencfv 
        - [x] diagnose `open diagnostic-d972998ab23145fa99258d883c204f21.zip: no such file or directory` ? Please someone verify that this is an actual error and not something that I might be doing wrong @karencfv @marclop fixed in #110.
        - [x] instances override-capacity @karencfv 
        - [x] instances pause @karencfv 
        - [x] instances resume @karencfv 
        - [x] instances start-routing @karencfv 
        - [x] instances stop-routing @karencfv 
        - [x] keystore set @karencfv 
        - [x] keystore show @karencfv 
        - [x] list @karencfv 
        - [x] monitoring disable @karencfv 
        - [x] monitoring enable @karencfv 
        - [x] plan cancel @karencfv 
        - [x] plan history @karencfv 
        - [x] plan monitor @karencfv 
        - [x] plan reapply @karencfv 
        - [x] plan update @karencfv 
        - [x] reallocate @karencfv 
        - [x] restart @karencfv 
        - [x] resync @karencfv 
        - [x] search @karencfv 
        - [x] show @karencfv 
        - [x] shutdown @karencfv 
        - [x] start @karencfv 
        - [x] upgrade @karencfv 
    - [x] kibana 
[...]

PR #114, opened 2019-12-30
cmd: Fix deployment notes

karencfv:
## Description
Fixes the previously broken `deployment note` commands.

With the new deployment api and the introduction of the "deployment resource" concept, the api calls to `/api/v1/deployments/{deployment_id}/notes` no longer work in the same manner. 

Until this api endpoint is updated to work using the deployment resource concept, the only deployment IDs it interacts with are the elasticsearch deployments related to the ES IDs (I tested this thoroughly using cURL directly as well as ecctl). So this means we cannot write notes on kibana, apm or appsearch for the time being. I left ~`// TODO:`~ comments above all related code[1]. In the UI all notes show up on the deployment's main page as well as the ES page.

The UI is using a different method, other than calls to this api endpoint, to leave comments on the deployments. This means that comments made through the UI will not show up on `deployment note list`.

Screen Shot 2019-12-31 at 11 50 24 AM

```console
$ go run main.go --config ce-aws deployment note list 85f25bd3c58548d284173a8099f4fbdb
TIME                       COMMENT ID                         USER    MESSAGE
2019-12-30T22:48:12.872Z   145f41577b874e46af187e54a7b24848   admin   ecctl - message
```

I changed the `--message` flag to `--comment`.  There is a global `--message` flag which prevented the flag's information from showing up in the `--help` output and the command docs.

I also refactored the files a bit by separating them into more files, adding multierror for validation etc.

[1] EDIT: I left normal comments above the code, as the linter doesn't like comments that start with `// TODO:` 😐 

## Related Issues
Closes: https://github.com/elastic/ecctl/issues/99

## How Has This Been Tested?
Manually:

```console
$ go run main.go --config ce-aws deployment note create 85f25bd3c58548d284173a8099f4fbdb --comment "ecctl - message"
$ go run main.go --config ce-aws deployment note list 85f25bd3c58548d284173a8099f4fbdb
TIME                       COMMENT ID                         USER    MESSAGE
2019-12-30T22:48:12.872Z   145f41577b874e46af187e54a7b24848   admin   ecctl - message
$ go run main.go --config ce-aws deployment note show 85f25bd3c58548d284173a8099f4fbdb --id 145f41577b874e46af187e54a7b24848
{
  "id": "145f41577b874e46af187e54a7b24848",
  "message": "ecctl - message",
  "timestamp": "2019-12-30T22:48:12.872Z",
  "user_id": "admin"
}
$ go run main.go --config ce-aws deployment note update 85f25bd3c58548d284173a8099f4fbdb --id 145f41577b874e46af187e54a7b24848 --comment "ecctl - message updated"
```

## Types of Changes

- [x] Refactoring (improves code quality but has no user-facing effect)
- [ ] Documentation

## Readiness Checklist


pkg/deployment/note/list.go
```diff
@@ -0,0 +1,68 @@
...
+	esID, err := getElasticsearchID(deployment.GetParams{
+		API:          params.API,
+		DeploymentID: params.ID,
+	})
+	if err != nil {
+		return err
+	}
+
+	params.ID = esID
+	return err
```
marclop: It's preferable to return the explicit nil since above `err != nil ` is already checked for. Otherwise always returning `err` without checking if `err != nil` is also less code

```suggestion
	return nil
```

pkg/deployment/note/show.go
```diff
@@ -21,24 +21,28 @@ import (
...
+}
 
-	res, err := params.API.V1API.Deployments.GetDeploymentNotes(
-		deployments.NewGetDeploymentNotesParams().
-			WithDeploymentID(params.ID),
-		params.AuthWriter,
-	)
+// Use different resource types when this is supported by the API.
+// For the time being, the notes endpoint only allows elasticsearch IDs.
+func (params *GetParams) fillDefaults() error {
```
marclop: we're doing this same workflow in a few places. I think it'd be better to tie the fillDefaults to a `NoteParams` and then embed that or just use it wherever needed.

cmd/deployment/note/command.go
```diff
@@ -73,16 +75,18 @@ var deploymentNoteListCmd = &cobra.Command{
 }
 
 var deploymentNoteUpdateCmd = &cobra.Command{
-	Use:     "update <deployment id> --id <note id> --message <message content>",
+	Use:     "update <deployment id> --id <note id> --comment <comment content>",
 	Short:   "Updates the deployment notes",
 	PreRunE: cmdutil.MinimumNArgsAndUUID(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
+		msg, _ := cmd.Flags().GetString("comment")
```
ppapapetrou76: minor : I'd rename the variable to `comment` - completely optional

cmd/deployment/note/command.go
```diff
@@ -38,16 +38,18 @@ var Command = &cobra.Command{
 }
 
 var deploymentNoteCreateCmd = &cobra.Command{
-	Use:     "create <deployment id> --message <message content> --type [elasticsearch|kibana|apm]",
+	Use:     "create <deployment id> --comment <comment content>",
 	Aliases: []string{"add"},
 	Short:   "Adds a note to a deployment",
 	PreRunE: cmdutil.MinimumNArgsAndUUID(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
+		msg, _ := cmd.Flags().GetString("comment")
```
ppapapetrou76: Same comment about the variable name

pkg/deployment/note/params.go
```diff
@@ -20,20 +20,17 @@ package note
...
+	multierror "github.com/hashicorp/go-multierror"
+
 	"github.com/elastic/ecctl/pkg/deployment"
 )
 
 var (
-	errEmptyNoteMessage = errors.New("note message cannot be empty")
-	errEmptyUserID      = errors.New("user id cannot be empty")
-	errEmptyNoteID      = errors.New("note id cannot be empty")
+	errEmptyNoteMessage = "note message cannot be empty"
```
ppapapetrou76: Again I prefer consistency so if  we plan to keep `comment` as the term our user are seeing then I'd use the same term all over the place :)

pkg/deployment/note/params.go
```diff
@@ -42,38 +39,23 @@ type Params struct {
...
-		return errEmptyNoteMessage
-	}
-
-	if params.UserID == "" {
-		return errEmptyUserID
-	}
-
-	if params.Params.NoteID == "" {
-		return errEmptyNoteID
+func getElasticsearchID(params deployment.GetParams) (string, error) {
```
ppapapetrou76: Does this function belong to the `note` package?

pkg/deployment/note/show.go
```diff
@@ -21,24 +21,28 @@ import (
...
+		DeploymentID: params.ID,
+	})
 	if err != nil {
-		return nil, api.UnwrapError(err)
+		return err
 	}
 
-	return res.Payload, nil
+	params.ID = esID
+	return err
```
ppapapetrou76: I assume we want to return `nil` here as per @marclop [here](https://github.com/elastic/ecctl/pull/114/files#r362165257)

pkg/deployment/note/show_test.go
```diff
@@ -0,0 +1,157 @@
...
+			args: args{
+				params: GetParams{
+					Params: Params{
+						Params: deployment.Params{},
+					},
+				},
+			},
+			wantErr: &multierror.Error{Errors: []error{
+				errors.New("note id cannot be empty"),
+				util.ErrAPIReq,
```
ppapapetrou76: I'd use the plan text here instead of the utility constant. The reason is that if accidentally we change the `ErrAPIReq` value to something that doesn't make sense the test will keep succeeding while it shouldn't 😄

pkg/deployment/note/update.go
```diff
@@ -0,0 +1,96 @@
...
+
+	merr = multierror.Append(merr, params.Params.Validate())
+
+	return merr.ErrorOrNil()
+}
+
+// Use different resource types when this is supported by the API.
+// For the time being, the notes endpoint only allows elasticsearch IDs.
+func (params *UpdateParams) fillDefaults() error {
+	esID, err := getElasticsearchID(deployment.GetParams{
```
ppapapetrou76: The same code appears 3 times. I wonder if we want to extract it in a single common function

pkg/deployment/note/update_test.go
```diff
@@ -0,0 +1,165 @@
...
+			args: args{
+				params: UpdateParams{
+					Params: Params{},
+				},
+			},
+			wantErr: &multierror.Error{Errors: []error{
+				errors.New("user id cannot be empty"),
+				errors.New("note message cannot be empty"),
+				errors.New("note id cannot be empty"),
+				util.ErrAPIReq,
```
ppapapetrou76: same comment here about using a constant in tests

pkg/deployment/note/write.go
```diff
@@ -19,58 +19,55 @@ package note
...
-		err = multierror.Append(err, errors.New("message cannot be empty"))
+	merr = multierror.Append(merr, params.Params.Validate())
+
+	return merr.ErrorOrNil()
+}
+
+// Use different resource types when this is supported by the API.
+// For the time being, the notes endpoint only allows elasticsearch IDs.
+func (params *AddParams) fillDefaults() error {
+	esID, err := getElasticsearchID(deployment.GetParams{
```
ppapapetrou76: 4th time :)

pkg/deployment/note/list.go
karencfv: 😄 I have no idea why I did that

karencfv:
@ppapapetrou76 @marclop thanks for reviewing! I've made the required changes :)

PR #117, opened 2020-01-02
release: v1.0.0-beta1 release fixes

marclop:
## Description

This patch fixes a slew of broken things ranging from unsupported brew
options to missing ldflags and S3 binary uploads.

* Updates the Goreleaser image to Go 1.13.x
* Updates the Goreleaser version itself to v0.123.3
* Removes the Brew section since we're not updating binaries via
goreleaser renders the automatic brew formula generation unusable.
* Adds missing BUILT ldflag to .goreleaser.yml
* Modifies the goreleaser config so it doesn't upload artifacts GitHub
  or create a GitHub release at all
* Corrects the v1.0.0-beta1.md release notes.
* Updates `scripts/goreleaser` so that it uploads binaries and checksums
  to AWS S3, and publish the GitHub release via `hub`.

## Motivation and Context

Kind of broken release

## How Has This Been Tested?

`make release`

## Types of Changes


build/Makefile.build
```diff
@@ -44,8 +45,9 @@ release:
 ifneq ($(SKIP_TAG),true)
 	@ $(MAKE) tag
 endif
-	@ echo "-> Releasing $(BINARY) v$(VERSION)..."
-	@ ./scripts/goreleaser release --rm-dist --release-notes=notes/$(VERSION).md
+	@ echo "-> Releasing $(BINARY) $(VERSION)..."
+	@ okta-awscli
```
ppapapetrou76: I don't think this is something our users will have installed :D

scripts/goreleaser
```diff
@@ -22,6 +24,17 @@ docker build -f ./build/Dockerfile.goreleaser -t ecctl:goreleaser .
...
     -e OWNER="${OWNER}" -e REPO="${REPO}" -e GO111MODULE=on \
     -e GITHUB_TOKEN="${GITHUB_TOKEN}" -e VERSION="${VERSION}" \
-    -v "$GOPATH:/go" \
+    -e BUILT="${BUILT}" -v "$GOPATH:/go" \
     -v "$(pwd):/go/src/github.com/${OWNER}/${REPO}" \
     -w "/go/src/github.com/${OWNER}/${REPO}" ecctl:goreleaser ${*}
+
+# Upload the binaries and the checksums.
+for f in dist/*.{tar.gz,deb,rpm,txt}; do
+    aws --profile ecsecurity s3 cp --acl bucket-owner-full-control \
```
ppapapetrou76: not sure if we want to expose these. I mean these don't mean anything to people outside `elastic`. WDYT?

marclop: We (the maintainers) are the only ones who are going to be releasing this and it's our profile, which is only a name here, not sure there's any harm

build/Makefile.build
marclop: We (the maintainers) are the only ones who are going to be releasing this and it's needed to obtain the necessary credentials. If we move to GitHub actions we can look into removing it.

scripts/goreleaser
ppapapetrou76: Completely harmless - just wondering if we want to follow this path 😄

notes/v1.0.0-beta1.md
```diff
@@ -152,131 +148,67 @@ The command now works as expected.
 <!-- AUTOGENERATED COMMIT LIST -->
 
 97c3985 Adding ecctl icon to repo (#111)
```
karencfv: The changes look like this in markdown, are we OK without the newlines?

Screen Shot 2020-01-03 at 10 07 05 AM

scripts/goreleaser
marclop: ah - of course, from my side I was curious as to if I had missed something

notes/v1.0.0-beta1.md
marclop: That's very strange. My local parser returns it as it's been published in the release, and also the release notes has been published with the same format as this PR publishes. Not sure why GitHub preview returns it as we see it.

karencfv: Looks like it's always shown like this in the [previous releases](https://github.com/elastic/cloud-cli/blob/master/notes/1.0.0-bc1.md#detailed-changes) 🤔 Do we want to keep it like this?

marclop:
@ppapapetrou76@karencfv I've changed the format to use bullet points so that doesn't happen anymore. Next releases will have this format.

PR #133, opened 2020-01-10
Updated instructions to use elastic/tap

zanbel:
Replacing previous PR since I forgot to sign the commit .. :/

I've also noticed master was not fully aligned with `v1.0.0-beta1` branch so aligned as part of this commit. Happy to split to a different PR if needed.


zanbel:
Thanks for the quick review! Anyone with write can you please merge.

PR #123, opened 2020-01-09
Add beta1 release notes

nrichers:
## Description

This PR adds release notes for 1.0.0-beta1 in a new "Release notes" section to our official ecctl documentation.

Content is based on https://docs.google.com/document/d/1nUq8cCiuMingMxYU-GvPojyuDi94jXieuh_oMonQzd0/edit# with some edits.

## Related Issues

Closes #26 

## How Has This Been Tested?

Content builds cleanly. Output:

image

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (improves code quality but has no user-facing effect)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation

## Readiness Checklist

- [ ] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [x] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #137, opened 2020-01-12
release: Automate release in GitHub action

marclop:
## Description

Automates the release process for ecctl binaries in a GitHub action. As
part of automating the release, it has been split into two parts;
running goreleaser and running goreleaser post actions.

Since goreleaser no longer automates the homebrew formula creation or
the upload of binaries to an artifact repository, those actions have
been moved to `scripts/goreleaser-post-actions.sh`, which uploads the
binaries to S3, Creates GitHub release, and updates elastic/homebrew-tap
with the latest ecctl version, binary links and checksums.

## Motivation and Context

Fully automated release process. No more user-tied releases.

## How Has This Been Tested?

Manually

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,55 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - v*.*.*
+
+jobs:
+  name: Release a new version
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Set up Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        id: go
+
+      - name: Install hub
+        uses: geertvdc/setup-hub@master
+
+      - name: Update AWS cli
+        uses: chrislennon/action-aws-cli@v1.1
+
+      - name: Get the version
+        id: get_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+
+      - name: Get the builtvar
+        id: get_built
+        run: echo ::set-output name=BUILT::$(date -u +%a_%d_%b_%H:%M:%S_%Y)
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_EXTRA }}
+          VERSION: ${{ steps.get_version.outputs.VERSION }}
+          BUILT: ${{ steps.get_built.outputs.BUILT }}
+          OWNER: elastic
+          REPO: ecctl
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - v*.*.*
```
karencfv: In relation to this [comment](https://github.com/elastic/ecctl/pull/123#discussion_r364861327) are we going to keep the `v` prefix?

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+          REPO: ecctl
+        with:
+          version: latest
+          args: release --rm-dist
+
+      - name: Run release post actions
+        run: ./scripts/goreleaser-post-actions.sh $VERSION
+        env: 
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_EXTRA }}
+          GITHUB_USER: marclop
```
karencfv: Is it possible to use `elastic`, the tooling team or each of us use our own usernames?

scripts/update-brew-tap.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Parameters
+VERSION=$(echo ${1}| sed 's/^v//')
+
+# Globals
+FORMULA_FILE=/tmp/homebrew-tap/Formula/ecctl.rb
+GITHUB_USER=marclop
```
karencfv: same as above

.github/workflows/release.yml
marclop: So, the trouble here is that the user name needs to match the username which created the `GH_TOKEN_EXTRA` which is used to fork and create the Pull Request too, that's why. 

In time I think a user with permissions to our repos which can be used for these sort of things is best.

This is a small step towards that direction

scripts/update-brew-tap.sh
marclop: Same as above

.github/workflows/release.yml
marclop: The v prefix is coming from the semver specification. We can keep it or just ignore it and on every tag with `*. *. *` if that's better

.github/workflows/release.yml
ppapapetrou76: I have the same issue in the soteria project where I'm using a personal token with extra permissions

.github/workflows/release.yml
karencfv: > just ignore it and on every tag with *.*.* if that's better

This sounds good to me if everyone's on board :)

marclop:
@ppapapetrou76 @karencfv I changed the tag format in 0ba906c

Issue #160, opened 2020-02-03
[ESS API Support] Hide some platform commands from help


Kushmaro:
When working with ecctl against ESS, we should hide some of the platform commands in the help section, at least.

## commands that should be hidden
- platform allocator
- platform constructor
- platform enrollment-token
- platform info 
- platform instance-configuration
- platform proxy
- platform repository
- platform role
- user (no user mgmt in ess currently)

## commands that should be kept
- platform deployment-template list/show 
- platform stack list/show

Autocomplete command would still return all the above but should return 401 if tried

ppapapetrou76:
@Kushmaro this is something that is brought for the first time to our attention. Can you please elaborate a bit more? I mean I understand the issue but users won't be able to get information from those commands. They will get `401`

What you are describing actually is to deliver a different version of `ecctl` for `ess`

Kushmaro:
> @Kushmaro this is something that is brought for the first time to our attention. Can you please elaborate a bit more? I mean I understand the issue but users won't be able to get information from those commands. They will get `401`
> 
> What you are describing actually is to deliver a different version of `ecctl` for `ess`

Discussed this one with @marclop , he'll check if its possible. 
if loe isn't high, we should do it IMO to improve ecctl<->ESS usage experience
if loe is high, we could settle for returning 401s

PR #169, opened 2020-02-07
cmd: improve help wording for ECE specific commands

karencfv:
## Description
Adds "(for ECE installations only)" to the descriptions of the commands which are ECE-specific.

Concerns were raised about hiding commands based on host address, our documentation, the fact that our code is open sourced so users will be able to see our commands anyway and may wonder why they cannot see them. 

Modifying 401 error messages to be ESS specific also brings in a fair bit of complexity.

I believe that in order to avoid complexity and still provide a good UX, the best course of action is to let the users know which commands will only work in ECE through the --help flag, and the documentation. Users will receive a 401 error if they attempt to run this command.

## Related Issues
Closes: https://github.com/elastic/ecctl/issues/160

## Types of Changes

- [x] Documentation

## Readiness Checklist

- [ ] I have added tests to cover my changes

### DIFF
--- a/cmd/platform/allocator/command.go
+++ b/cmd/platform/allocator/command.go
@@ -26,7 +26,7 @@ import (
 // Command represents the allocator command
 var Command = &cobra.Command{
 	Use:     "allocator",
-	Short:   "Manages allocators",
+	Short:   "Manages allocators (for ECE installations only)",
 	PreRunE: cobra.MaximumNArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		cmd.Help()
--- a/cmd/platform/constructor/command.go
+++ b/cmd/platform/constructor/command.go
@@ -36,7 +36,7 @@ const (
 // Command represents the constructor command
 var Command = &cobra.Command{
 	Use:     "constructor",
-	Short:   "Manages constructors",
+	Short:   "Manages constructors (for ECE installations only)",
 	PreRunE: cobra.MaximumNArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		cmd.Help()
--- a/cmd/platform/enrollment-token/command.go
+++ b/cmd/platform/enrollment-token/command.go
@@ -35,7 +35,7 @@ const (
 // Command represents the enrollment-token subcomand.
 var Command = &cobra.Command{
 	Use:     "enrollment-token",
-	Short:   "Manages tokens",
+	Short:   "Manages tokens (for ECE installations only)",
 	PreRunE: cobra.MaximumNArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		cmd.Help()
--- a/cmd/platform/instance-configuration/command.go
+++ b/cmd/platform/instance-configuration/command.go
@@ -32,7 +32,7 @@ import (
 // Command is the top instance-config subcommand.
 var Command = &cobra.Command{
 	Use:     "instance-configuration",
-	Short:   "Manages instance configurations",
+	Short:   "Manages instance configurations (for ECE installations only)",
 	PreRunE: cobra.MaximumNArgs(0),
 	Run:     func(cmd *cobra.Command, args []string) { cmd.Help() },
 }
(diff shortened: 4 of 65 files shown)


zanbel:
After discussing today with @Kushmaro wanted to share my $.02 although I see the PR is merged but still maybe something to consider. The indication added about specific endpoint only being relevant for ECE are a bug, IMO. It's not accurate to say `(for ECE installations only)` since the commands are relevant for both ESS and ECE but require certain permissions. If we follow the same logic we should add indication for user with deployments manager or viewer roles about certain endpoints being `for platform admins only`.

My suggestion will be to let the API return a 403 in case the api key presented does not have enough permissions. And one more point about UI vs. API/CLI, in the user console we have an active session, meaning we can more easily customize the experience based on the authentication presented by the user. In API/CLI it's harder to assume that. Specifically with ecctl if API key is configured in the ecctl.config we can potentially try to fetch this information before every command by I have not seen this happened in similar tools. if there is an example of such practice I'll be happy to see it.

@Kushmaro @ppapapetrou76 @marclop @karencfv -- I'm happy to assist in a design discussion.

// cc @elastic/cloud-pms if folks might have other opinions or saw different examples

karencfv:
@zanbel you make a very good point about the different roles. I'm perfectly fine with removing the extra text added in this PR, and letting the API return a 403. It would also be a good idea to mention this in our documentation.

About fetching roles from the user before every command, even though it's possible, I don't think it's a course of action we should pursue. Making another API call before every command would not only make the CLI slower, but would IMO add unnecessary complexity to our code.

marclop:
Thanks for chiming in Roy, I think something along the lines of: `(Requires platform administration privileges)` might be a bit more precise, we can certainly change the working here if you're happy with that.

I'm also of the opinion that adding 1 extra API call can make the CLI seem slower than necessary since it would become an API call which needs to get made before the real command execution begins, that would require us to retrieve the permissions which the user has and hide or show specific parts of the help prompt. I think that adds a certain complexity which I'm not sure it's worth investing on now.

Regardless of the help text, the API will return a 403 (Maybe even a 404) if the user doesn't have the appropriate permissions to access those endpoints.

I have not seen other projects dynamically adjusting their help like you say, but it's something we might be able to do, but it can slow down the user's experience.

One thing we might want to do if we decided to go the route of fetching the API key's information and tied roles is to cache that information in the filesystem (with the checksum of the APIKey to make sure we don't return outdated information to the user if the APIKey changes) to reduce the cost of making an API call. This however will further increase the complexity of that workflow.

karencfv:
Thanks for expanding on that Marc :) I like the text you propose.

I'd also like to add that if we decided to go with the role based customised experience, would that mean that we'd have to customise for each role (platform admin, platform viewer, deployments manager, and deployments viewer)? I genuinely believe this could get pretty complex and  error prone.

ppapapetrou76:
👎 to the extract API call
👍 to leave the command labels/wording common across all targeting envs
👍 to replace the `403` or `404` with something more user-friendly 

My 2 cents but I'll leave @marclop and @karencfv make the call

Kushmaro:
Thanks for the fix @karencfv 

> Concerns were raised about hiding commands based on host address, our documentation, the fact that our code is open sourced so users will be able to see our commands anyway and may wonder why they cannot see them.

As I've mentioned, modifying the help command was more about helping the simple user know which commands they have available for use, not about hiding anything else really.
Keep in mind that user who reads our open source isn't necessarily the user would use the cli (i'm looking more towards an administrative sort of persona rather a dev)

given this approach, I'd also prefer "Requires administrative access" rather "for ece installation only", If we look at the *majority* of the users who we intend ecctl for, they are 'simple' dev or admin users, with no admin rights, and would never have such admin rights.

internal users - should be a bit smarter than the common external user since they have some internal knowledge of the works & environments.

Which is why I personally find this path, generally less desirable.

zanbel:
I have to leave one more comment here since :) 
Would recommend looking for inspiration in other tools you use and like, and explore how they are handling similar scenarios. This is a very common use case and aligning with commonly used CLI tool can more easily back up any proposal instead of "inventing" something new. You probably have a handy list of CLI tools you often use and have an opinion about how happy you are when using them.

ppapapetrou76:
Folks not sure if a merged / closed PR is the best place to keep discussing about it :)

Issue #159, opened 2020-02-03
[ESS API support] Default ecctl URL configuration to ESS API


Kushmaro:
## ecctl should default to ESS
When initializing ecctl, we should ask the user for an API url, but default to the ess Public API

### Init command text
From: “Enter the API URL of your ECE installation”
To: “Enter the custom API URL for your Elastic Cloud Enterprise or Elastic Cloud environment, or hit enter to use Elasticsearch Service by default”

### Following configurations
- If the user opts to use ESS, we should suggest API authentication only as suggested in #157 
- If the user opts to use ESS, we should ask for the default region to work against #162

zanbel:
Wanted to add that I'm on the fence about not requiring users to explicitly choosing between ECE and ESS since I think it will reduce confusion and will treat both in the same way. E.g. we can add the following step:
```
Select if this configuration will be used with:
  [1] Elastic Cloud Enterprise
  [2] Elasticsearch Service // (or [2] Elastic Cloud)
```

Another thing to keep in mind is whether the term used should be `Elastic Cloud` or `Elasticsearch Service`. Currently ecctl only allows users to interact with the Elasticsearch Service, but even if we decide to go with the more future proof option, which is fine, please make sure we don't use `Elastic Cloud service`.

Kushmaro:
@zanbel as mentioned in the email thread, I'd still want to default to ESS in the URL section, without an explicit choice, as most users will be ESS users.

“Enter the custom API URL for your Elastic Cloud Enterprise or Elastic Cloud environment, or hit enter to use Elasticsearch Service by default”:

karencfv:
Thank you both for the input for the workflow 🙂  I have the following to add:

I agree with @Kushmaro that the default should be the ESS api host address and I also agree with @zanbel about the format of the step to avoid confusion. Having the step formatted in that way, apart from being consistent with the rest of the workflow, allows us more flexibility in the future to add ESSP, GovCloud etc.

I have already made a PR with the changes, to get this issue moving along and we can make adjustments as required.

Issue #162, opened 2020-02-03
[ESS API Support] Handle regions in config when ESS is configured


Kushmaro:
## Original issue
In case a user opts to use ESS in the init flow (#159), we should query for a default region and handle it in the ecctl configuration file like we did in the pre-beta1 release

when configured with ece, we should either just used a default non-meaningful region, or ignore this flag completely 

## Update

We chatted about this yesterday and agreed that the desired thing would be to query an API endpoint and print a list of available regions, said endpoint is not available today so in the meantime maintaining a static list of providers is low effort and a quick win, any time a new release of ecctl is shipped, new regions can be included which makes it a valid distribution mechanism.

zanbel:
Want to expand the scope of the issue a bit and also add the following.
1. Region config field / param: I think this is what Omer is suggestions, but just to make it more explicit, I would suggest to always have a region field in the configuration but in case of ECE it should set to "ece" and behavior should be consistent if the user will use a `--region ece`. Reason this is important to allow users to submit `ecclt deployments list --host $MY_ECE_URL --region ece` even when a default region is configured. 
2. Global vs. regional resources: even if configure to work with ESS API some resources will be global. For example certain account level configuration such as traffic filter rules. We need to identify which resources are global and which are regions and build this logic to be easily configurable and discoverable.

marclop:
We chatted about this yesterday and agreed that the desired thing would be to query an API endpoint and print a list of available regions, said endpoint is not available today so in the meantime maintaining a static list of providers is low effort and a quick win, any time a new release of ecctl is shipped, new regions can be included which makes it a valid distribution mechanism.

PR #173, opened 2020-02-10
cmd: init ask for type of infrastructure and default to ESS config setup

karencfv:
## Description
The init command will now ask for the type of infrastructure to determine whether to ask for a custom host address or to use the default ESS public api address (https://api.elastic-cloud.com) .

The output is as follows:

When choosing ESS:
```console
Select which type of infrastructure this configuration will for:
  [1] Elastic Cloud (Elasticsearch Service).
  [2] Elastic Cloud Enterprise (ECE).

Please enter your choice: 1

What default output format...
```

When choosing ECE:
```console
Select which type of infrastructure this configuration will for:
  [1] Elastic Cloud (Elasticsearch Service).
  [2] Elastic Cloud Enterprise (ECE).

Please enter your choice: 2

Enter the URL of your ECE installation:
```

Default to https://api.elastic-cloud.com:
```console
Select which type of infrastructure this configuration will for:
  [1] Elastic Cloud (Elasticsearch Service).
  [2] Elastic Cloud Enterprise (ECE).

Please enter your choice: 5

invalid choice, defaulting to "https://api.elastic-cloud.com"
```

## Related Issues
Closes: https://github.com/elastic/ecctl/issues/159

## Motivation and Context
Improved UX now that ESS public api will be available

## How Has This Been Tested?
Manually and unit tests

## Types of Changes

- [ ] Documentation

## Readiness Checklist

- [ ] My change requires a change to the documentation

### DIFF
--- a/pkg/ecctl/init.go
+++ b/pkg/ecctl/init.go
@@ -49,6 +49,12 @@ const (
 )
 
 const (
+	_ = iota
+	essInfraChoice
+	eceInfraChoice
+)
+
+const (
 	disclaimer      = "Welcome to the Elastic Cloud CLI! This command will guide you through authenticating and setting some default values.\n\n"
 	redacted        = "[REDACTED]"
 	settingsPathMsg = "Found existing settings in %s. Here's a JSON representation of what they look like:\n"
@@ -67,6 +73,13 @@ const (
 )
 
 var (
+	hostChoiceMsg = `
+Select which type of infrastructure this configuration will for:
+  [1] Elastic Cloud (Elasticsearch Service).
+  [2] Elastic Cloud Enterprise (ECE).
+
+Please enter your choice: `
+
 	authChoiceMsg = `
 Which authentication mechanism would you like to use?
   [1] API Keys (Recommended).
@@ -186,7 +199,10 @@ func InitConfig(params InitConfigParams) error {
 	// Insecure is set to true by default to allow API calls against HTTPS
 	// endpoints with self-signed certificates.
 	cfg.Insecure = true
-	cfg.Host = scanner.Scan(hostMsg)
+
+	if err := askInfraSelection(&cfg, scanner, params.Writer, params.ErrWriter); err != nil {
+		return err
+	}
 
 	if err := askOutputFormat(&cfg, scanner, params.Writer, params.ErrWriter); err != nil {
 		return err
@@ -206,7 +222,7 @@ func InitConfig(params InitConfigParams) error {
 		return err
 	}
 
-	// It's better to write the config as is since it ommits defaults and
+	// It's better to write the config as is since it omits defaults and
 	// empties vs viper's behaviour in `WriteConfig`.
 	return writeConfig(cfg, params.FilePath, ".json")
 }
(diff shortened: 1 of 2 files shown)


pkg/ecctl/init.go
```diff
@@ -243,6 +259,26 @@ func printConfig(writer io.Writer, v *viper.Viper) error {
...
+	infraChoice, err := strconv.Atoi(infraChoiceRaw)
+	if err != nil {
+		return err
+	}
+
+	cfg.Host = "https://api.elastic-cloud.com"
+	switch infraChoice {
+	case eceInfraChoice:
+		cfg.Host = scanner.Scan(hostMsg)
+	case essInfraChoice:
```
marclop: Maybe we can print a line stating the URL which will be used? It seems a bit empty when you select that you get no confirmation given that the other 2 options (Invalid or ECE, do have some more info).

pkg/ecctl/init.go
```diff
@@ -243,6 +259,26 @@ func printConfig(writer io.Writer, v *viper.Viper) error {
...
 
+func askInfraSelection(cfg *Config, scanner *input.Scanner, writer, errWriter io.Writer) error {
+	infraChoiceRaw := scanner.Scan(hostChoiceMsg)
+	fmt.Fprintln(writer)
+	infraChoice, err := strconv.Atoi(infraChoiceRaw)
+	if err != nil {
+		return err
+	}
+
+	cfg.Host = "https://api.elastic-cloud.com"
```
marclop: I'd make this a constant to use it here and in line 276 too.

marclop:
I'm also wondering if we should prompt for a region when ESS is select and wether or not to provide a sane default.

karencfv:
@marclop sounds good! we'll add that with https://github.com/elastic/ecctl/issues/162

pkg/ecctl/init.go
```diff
@@ -67,6 +73,13 @@ const (
 )
 
 var (
+	hostChoiceMsg = `
```
ppapapetrou76: I'm sure you already thought about it but do the `govcloud` and `essp` env require a different approach during our init flow?

pkg/ecctl/init.go
```diff
@@ -67,6 +73,13 @@ const (
...
+Select which type of infrastructure this configuration will for:
+  [1] Elastic Cloud (Elasticsearch Service).
```
ppapapetrou76: I'd like @Kushmaro to confirm the wording here. It seems that the official name is `Elasticsearch Service` so imho this should be the name not inside the parenthesis, but I might be wrong

pkg/ecctl/init.go
```diff
@@ -48,6 +48,12 @@ const (
...
 )
 
+const (
+	_ = iota
+	essInfraChoice
+	eceInfraChoice
+)
+
 const (
 	disclaimer      = "Welcome to the Elastic Cloud CLI! This command will guide you through authenticating and setting some default values.\n\n"
```
Kushmaro: for ecctl we use the phrase "Elastic Cloud Control (ecctl)" rather Elastic Cloud CLI

pkg/ecctl/init.go
Kushmaro: thanks @ppapapetrou76 for the mention!
@karencfv , I'd change the wording to - 
"Select which type of Elastic Cloud offering will you be working with:"
[1] Elasticsearch Service (default)
[2] Elastic Cloud Enterprise (ECE) or Elasticsearch Service Private (ESSP)

 if no choice is enter - [1] should be selected.
in case 1 is selected, we could also output something like - 
"Using https://api.elastic-cloud.com as the API host" 

and continue the flow

zanbel:
>I'm also wondering if we should prompt for a region when ESS is select and whether or not to provide a sane default too.

Based on previous research down as part of randomizing provide and region, I don't think there can be a sane default here. Users are very selective about their region of choice, even if they don't have a strong preference about which cloud provider to use.

>Select which type of Elastic Cloud offering will you be working with:"
[1] Elasticsearch Service (default)
[2] Elastic Cloud Enterprise (ECE) or Elasticsearch Service Private (ESSP)

Just to mention that, per the other [comment](https://github.com/elastic/ecctl/issues/159#issuecomment-583840820), you might want to actually offer 3 options and the reason is that, AFAIK, the way the URL is constructed in ESSP is similar to ESS, not ECE, and related to the need to include specific region. If that's not the case two options can be fine.

Kushmaro:
> Just to mention that, per the other [comment](https://github.com/elastic/ecctl/issues/159#issuecomment-583840820), you might want to actually offer 3 options and the reason is that, AFAIK, the way the URL is constructed in ESSP is similar to ESS, not ECE, and related to the need to include specific region. If that's not the case two options can be fine.

sure, that's up for @marclop @karencfv to comment implementation wise, my reasoning for 2 options was that for the foreseeable timeframe, ECE / ESSP should behave pretty much the same so didn't see a real reason to duplicate the flow

If you feel 3 choice is a bit more future proof in terms of ESSP - sure lets do that

pkg/ecctl/init.go
karencfv: I had thought to change this afterwards but might as well do it now :)

karencfv:
Thanks everyone for your comments 🙇‍♀ I've updated the PR with the required changes

PR #174, opened 2020-02-11
cmd: improve UX by defining which commands are admin specific

karencfv:
## Description
This PR introduces the text "(Requires platform administration privileges)" to those commands that require it. This improves UX for ESS users. 

## Related Issues
Related: https://github.com/elastic/ecctl/issues/160

## Motivation and Context
After a conversation in https://github.com/elastic/ecctl/pull/169 it was concluded that the best way forward was to use the text mentioned previously.

## Types of Changes

- [x] Documentation

## Readiness Checklist

- [ ] I have added tests to cover my changes

### DIFF
--- a/cmd/platform/allocator/command.go
+++ b/cmd/platform/allocator/command.go
@@ -18,8 +18,6 @@
 package cmdallocator
 
 import (
-	"fmt"
-
 	"github.com/spf13/cobra"
 
 	cmdallocatormetadata "github.com/elastic/ecctl/cmd/platform/allocator/metadata"
@@ -29,7 +27,7 @@ import (
 // Command represents the allocator command
 var Command = &cobra.Command{
 	Use:     "allocator",
-	Short:   fmt.Sprintf("Manages allocators %v", cmdutil.PlatformAdminRequired),
+	Short:   cmdutil.AdminReqDescription("Manages allocators"),
 	PreRunE: cobra.MaximumNArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		cmd.Help()
--- a/cmd/platform/allocator/command.go
+++ b/cmd/platform/allocator/command.go
@@ -26,7 +26,7 @@ import (
 // Command represents the allocator command
 var Command = &cobra.Command{
 	Use:     "allocator",
-	Short:   "Manages allocators (for ECE installations only)",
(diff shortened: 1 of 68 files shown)


karencfv:
Moving the conversation over to this PR. If any changes need to be addressed let's do it here :)

cmd/platform/allocator/command.go
```diff
@@ -26,7 +26,7 @@ import (
 // Command represents the allocator command
 var Command = &cobra.Command{
 	Use:     "allocator",
-	Short:   "Manages allocators (for ECE installations only)",
+	Short:   "Manages allocators (Requires platform administration privileges)",
```
marclop: Maybe we can define the text as a constant and use `fmt.Sprintf()` to format this.

cmd/platform/allocator/command.go
```diff
@@ -18,15 +18,18 @@
...
 
 	cmdallocatormetadata "github.com/elastic/ecctl/cmd/platform/allocator/metadata"
+	cmdutil "github.com/elastic/ecctl/cmd/util"
 )
 
 // Command represents the allocator command
 var Command = &cobra.Command{
 	Use:     "allocator",
-	Short:   "Manages allocators (for ECE installations only)",
+	Short:   fmt.Sprintf("Manages allocators %v", cmdutil.PlatformAdminRequired),
```
ppapapetrou76: So we are repeating the same naming pattern a few dozens of times and if - in the future - we will need to change it or improve we will need to do the same which is kinda error prone.
How about introduce a function that will accept the command `message` and wrap it with the permissions message. Something like (ignore the function and variable names - pick those that make more sense)
```go
func shortName(name string) string {
    return fmt.Sprintf("%s %v", name, cmdutil.PlatformAdminRequired)
}
```

WDYT?
This will also remove the `cmdutil` deps from all commands and will eliminate some ugly imports like `sdkcmdutil` 😄

marclop: I think that makes a lot of sense, Using a function with that 1 param works well.

Kushmaro:
Hey @karencfv 
I've discussed this one with @uric 

really sorry for the hassle, but could we please change the “Requires platform administration privileges” to “Available on ECE only”

We believe this would make more sense from an ESS user's perspective

PR #180, opened 2020-02-14
cmd: init gives you a choice to select default region when ESS is selected

karencfv:
## Description
Whenever a user opts to use ESS in the init flow, they will be prompted to select a default region.

```console
Select which type of Elastic Cloud offering you will be working with:
  [1] Elasticsearch Service (default).
  [2] Elastic Cloud Enterprise (ECE).
  [3] Elasticsearch Service Private (ESSP).

Please enter your choice: 1

Using "https://api.elastic-cloud.com" as the API endpoint.

Select a region you would like to have as default:
  
  GCP
  [1] us-central1 (Iowa)
  [2] us-east4 (N. Virginia)
  [3] us-west1 (Oregon)
  [4] northamerica-northeast1 (Montreal)
  [5] australia-southeast1 (Sydney)
  [6] europe-west1 (Belgium)
  [7] europe-west2 (London)
  [8] europe-west3 (Frankfurt)
  [9] asia-northeast1 (Tokyo)
  [10] asia-south1 (Mumbai)

  AWS
  [11] us-east-1 (N. Virginia)
  [12] us-west-1 (N. California)
  [13] us-west-2 (Oregon)
  [14] eu-central-1 (Frankfurt)
  [15] eu-west-2 (London)
  [16] eu-west-1 (Ireland)
  [17] ap-northeast-1 (Tokyo)
  [18] ap-southeast-1 (Singapore)
  [19] ap-southeast-2 (Sydney)
  [20] sa-east-1 (São Paulo)

  Azure
  [21] eastus2 (Virginia)
  [22] westus2 (Washington)
  [23] westeurope (Netherlands)
  [24] japaneast (Tokyo)
  [25] southeastasia (Singapore)

Please enter your choice: 
```

Things to note: 

- ESSP defaults to us-west-2 as it's the only available region for the time being
- In order to have some consistency in the list, for the region names, I've decided to go with this format instead of the display format when available. This is because GCP does not have a display format like the other two.
- The desired thing would be to query an API endpoint and print a list of available regions, said endpoint is not available today so in the meantime maintaining a static list of providers is low effort and a quick win, any time a new release of ecctl is shipped, new regions can be included which makes it a valid distribution mechanism.

## Related Issues
Closes: https://github.com/elastic/ecctl/issues/162

## Motivation and Context
UX improvements

## How Has This Been Tested?
Manually and unit tests

## Types of Changes

- [ ] Documentation

## Readiness Checklist

- [ ] My change requires a change to the documentation

### DIFF
--- a/pkg/ecctl/init.go
+++ b/pkg/ecctl/init.go
@@ -355,6 +355,7 @@ func askInfraSelection(cfg *Config, scanner *input.Scanner, writer, errWriter io
 	return nil
 }
 
+//nolint
 func askRegionSelection(cfg *Config, scanner *input.Scanner, writer io.Writer) error {
 	regionChoiceRaw := scanner.Scan(regionChoiceMsg)
 	fmt.Fprintln(writer)
--- a/pkg/ecctl/init.go
+++ b/pkg/ecctl/init.go
@@ -56,6 +56,35 @@ const (
 )
 
 const (
+	_ = iota
+	gcpUsCentral1Choice
+	gcpUsEast4Choice
+	gcpUsWest1Choice
+	gcpNorthamericaNortheast1Choice
+	gcpAustraliaSoutheast1Choice
+	gcpEuropeWest1Choice
+	gcpEuropeWest2Choice
+	gcpEuropeWest3Choice
+	gcpAsiaNortheast1Choice
+	gcpAsiaSouth1Choice
+	awsUsEast1Choice
+	awsUsWest1Choice
+	awsUsWest2Choice
+	awsEuCentral1Choice
+	awsEuWest2Choice
+	awsEuWest1Choice
+	awsApNortheast1Choice
+	awsApSoutheast1Choice
+	awsApSoutheast2Choice
+	awsSaEast1Choice
+	azureEastUs2Choice
+	azureWestUs2Choice
+	azureWestEuropeChoice
+	azureJapanEastChoice
+	azureSouthEastAsiaChoice
+)
+
+const (
 	disclaimer      = "Welcome to Elastic Cloud Control (ecctl)! This command will guide you through authenticating and setting some default values.\n\n"
 	redacted        = "[REDACTED]"
 	settingsPathMsg = "Found existing settings in %s. Here's a JSON representation of what they look like:\n"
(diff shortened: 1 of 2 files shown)


pkg/ecctl/init.go
```diff
@@ -287,6 +355,73 @@ func askInfraSelection(cfg *Config, scanner *input.Scanner, writer, errWriter io
...
+//nolint
+func askRegionSelection(cfg *Config, scanner *input.Scanner, writer io.Writer) error {
+	regionChoiceRaw := scanner.Scan(regionChoiceMsg)
+	fmt.Fprintln(writer)
+	regionChoice, err := strconv.Atoi(regionChoiceRaw)
+	if err != nil {
+		return err
+	}
+
+	switch regionChoice {
```
marclop: We can replace all of this for `map[int]string` and then having a `region, ok := regionMap[regionChoice]` Handling the `ok` as the error and returning the region, It'll make the code shorter.

pkg/ecctl/init.go
```diff
@@ -55,6 +55,35 @@ const (
 	esspInfraChoice
 )
 
+const (
+	_ = iota
+	gcpUsCentral1Choice
```
ppapapetrou76: This is not realistic. With the pace regions are added we will need to update it very frequently . TBH It's already outdated :) there's a new region `us-east-2` coming really shortly
Additionally the first ESSP multi-region is coming so we need to keep that in mind as well.

pkg/ecctl/init.go
```diff
@@ -287,6 +355,73 @@ func askInfraSelection(cfg *Config, scanner *input.Scanner, writer, errWriter io
...
+func askRegionSelection(cfg *Config, scanner *input.Scanner, writer io.Writer) error {
+	regionChoiceRaw := scanner.Scan(regionChoiceMsg)
+	fmt.Fprintln(writer)
+	regionChoice, err := strconv.Atoi(regionChoiceRaw)
+	if err != nil {
+		return err
+	}
+
+	switch regionChoice {
+	case gcpUsCentral1Choice:
```
ppapapetrou76: Ouch

Kushmaro:
Product-wise, manually entering a region name should be fine as well IMO

karencfv:
HI all! Yes, I 100% agree that having a static list of providers is far from ideal. 

During the "ESS API blog post examples" meeting, the implementation of this feature came up. It was agreed that there is no available endpoint which allows us to query an API endpoint and print a list of available regions without authentication, this was the best way forward for the time being until we get such an endpoint (A note was left in the original issue about this conversation https://github.com/elastic/ecctl/issues/162#issuecomment-584953655). Please feel free to add anything from that meeting @marclop @zanbel :)

marclop:
@ppapapetrou76 We had a short meeting with Roy and Karen and discussed better on-boarding workflows for users, we agreed that ideally we'd have a list of regions to display to the user and that the most viable path forward was to have an API endpoint return the list of available regions which we'd return to the user.

Currently, the API endpoint is not available, and while having it really soon would be ideal it's unrealistic to expect such a fast turn around on it.

We agreed that while not ideal it's not a terrible solution to have a hardcoded list of regions somewhere, ecctl or cloud-sdk-go would be both good places where to have that list, i.e. `api.Regions` as a map and that could be consumable through ecctl, and we could add a helper function to it not have to handle the `region, ok := api.RegionMap`

karencfv:
Thanks for chiming in @marclop ! If we do decide to stick with the original plan and have a list of hard-coded regions I believe it'd be better to have them live in ecctl rather than cloud-sdk-go as it'll be easier to update/remove said list. I wouldn't want this temporary solution to have a longer life than it should 😄

ppapapetrou76:
Ok understood. Thanks for the background information. 
Can we please at least support `essp` and `govcloud` envs?

karencfv:
heya @ppapapetrou76 can you 👀 this again? I would rather not dismiss your "change requested" 😄

Issue #221, opened 2020-03-18
Cloud Saas example deployment wrong


111andre111:
## Expected Behavior

Deployment should be created without error.

## Current Behavior

The documented way of creating a new deployment may work for ECE, but it doesn't work with Cloud SaaS.
I found at least 2 references, maybe there exist more in the documentation:
https://www.elastic.co/guide/en/ecctl/current/ecctl_deployment_create.html
https://github.com/elastic/ecctl/blob/master/docs/ecctl_deployment_create.md#examples

`ecctl deployment create --file ./deployment_example.json --name=my-new-deployment --version=7.4.1`
It results in error message:
`Instance configuration [data.default] does not exist or is not available to you. The instance configuration must be one of the available configurations. Use the Instance Configuration API to get a list of available configurations.`

## Possible Solution

I did lots of investigation and came up with this solution creating a deployment with elasticsearch, kibana and apm:
```
cat deployment_example_gcp.json

{
  "cluster_name": "",
  "plan": {
    "cluster_topology": [
      {
        "node_type": {
          "data": true,
          "master": true,
          "ingest": true
        },
        "instance_configuration_id": "gcp.data.highio.1",
        "zone_count": 3,
        "size": {
          "resource": "memory",
          "value": 65536
        },
        "elasticsearch": {
          "enabled_built_in_plugins": []
        }
      },
      {
        "node_type": {
          "data": false,
          "master": false,
          "ingest": false,
          "ml": true
        },
        "instance_configuration_id": "gcp.ml.1",
        "zone_count": 1,
        "size": {
          "resource": "memory",
          "value": 16384
        },
        "elasticsearch": {
          "enabled_built_in_plugins": []
        }
      }
    ],
    "elasticsearch": {
      "version": "7.4.0"
    },
    "transient": {},
    "deployment_template": {
      "id": "gcp-io-optimized"
    }
  },
  "settings": {
    "dedicated_masters_threshold": 6
  },
  "kibana": {
    "plan": {
      "cluster_topology": [
        {
          "instance_configuration_id": "gcp.kibana.1",
          "zone_count": 1,
          "size": {
            "resource": "memory",
            "value": 1024
          }
        }
      ],
      "kibana": {
        "version": "7.4.0"
      }
    }
  }
}


ecctl deployment create --file ./deployment_example_gcp.json --name=my-new-deployment --version=7.4.1
```

## Context

I found at least another issue:

[...]

Issue #116, opened 2020-01-02
Add a flag-based deployment create


marclop:
## Overview

Add an easy way to create deployments other than via a file definition. Much like what was proposed in #67 bounded for `deployment elasticsearch create` which is now defunct.

See the markdown in: https://github.com/marclop/ecctl/blob/ae757de7c7694e2160e1443bf184c69c82bf997b/cmd/deployment/elasticsearch/create_help.go

## Possible Implementation

Using an intermediate structure much like the `--topology-element` flag in `elasticsearch create` with the ability to specify different resource types.

A boolean `--generate-payload` should be available so that the generated simplified topology can be saved into a file.

marclop:
Moving this to MS35 since higher priority tasks will be worked in

PR #224, opened 2020-03-23
cmd: add instance flags to deployment create command

karencfv:
## Description
Adds instance definition flags to the `ecctl deployment create` command so it is not necessary to use a predefined deployment definition JSON file.

More information about the command can be found in its help [documentation](https://github.com/karencfv/ecctl/blob/adc371243a3c01d26c26d9c95ff4884f4e807104/docs/ecctl_deployment_create.md).

**This is a breaking change since it changes the override behaviour the flags previously had.**

### Things to note

- All deployments must at least have one ES and one Kibana instance. This mimics the UI.

- For now each resource type that is being created makes a call to retrieve the deployment template definition. This is far from ideal, but the `deployment {kibana | apm | appsearch} create` commands use the same functions to update deployments with new instances. 

  Refactoring this falls out of the scope of this PR. I will write a separate issue to integrate the `deployment {kibana | apm | appsearch} create` commands into the `deployment update` command which will address this issue and have the `deployment update` command act in a similar way to the `deployment create` command with the newly introduced flags.

- I have added `params.DeploymentID != "" && len(params.DeploymentID) != 32` to the `NewStateless` validation as a temporary measure until the above has been implemented.

- I have removed all output examples from the help text that weren't strictly necessary as per @nrichers suggestion on a PR that I can't find any more 😄 (Nik, please correct me if I misunderstood).

- The links on the help documentation don't work yet, but they will once the ESS public API is released

## Related Issues
Closes: https://github.com/elastic/ecctl/issues/116
Closes: https://github.com/elastic/ecctl/issues/221

## Motivation and Context
UX improvement

## How Has This Been Tested?
Unit tests and manually testing against a ce-aws environment
 
## Types of Changes

- [ ] Documentation

## Readiness Checklist


pkg/deployment/depresource/new_stateless.go
```diff
@@ -30,7 +30,7 @@ import (
 type NewStateless struct {
 	*api.API
 
-	// Required: DeploymentID.
+	// Optional DeploymentID.
```
marclop: How come?

pkg/deployment/depresource/new_deployment.go
```diff
@@ -0,0 +1,135 @@
...
+
+		resources.Appsearch = []*models.AppSearchPayload{appsearchPayload}
+	}
+
+	payload := &models.DeploymentCreateRequest{
+		Name:      params.Name,
+		Resources: &resources,
+	}
+
+	return payload, nil
```
marclop:
```suggestion
	payload := models.DeploymentCreateRequest{
		Name:      params.Name,
		Resources: &resources,
	}

	return &payload, nil
```

pkg/deployment/depresource/new_deployment.go
```diff
@@ -0,0 +1,135 @@
...
+	Region                 string
+	ElasticsearchRefID     string
+	KibanaRefID            string
+	ApmRefID               string
+	AppsearchRefID         string
+	ElasticsearchSize      int32
+	ElasticsearchZoneCount int32
+	KibanaSize             int32
+	KibanaZoneCount        int32
+	ApmSize                int32
```
ppapapetrou76: I wonder if we can group the same per cluster type attributes in a structure to increase readability
I can see there are attributes like `size`, `zoneCount` and maybe `enable` that are repeated

pkg/deployment/depresource/new_stateless.go
karencfv: Because a new deployment doesn't have an ID yet 😄 

From what I wrote on the description ->
> I have added params.DeploymentID != "" && len(params.DeploymentID) != 32 to the NewStateless validation as a temporary measure until the above has been implemented.

This will be cleaned up when the `deployment update` flags issue is implemented.

pkg/deployment/depresource/new_deployment.go
```diff
@@ -24,46 +24,45 @@ import (
 	"github.com/elastic/cloud-sdk-go/pkg/models"
 )
 
+// InstanceParams holds the common instance fields for a resource
+type InstanceParams struct {
```
marclop: Why not follow the API naming? Maybe calling it `ResourceParams` makes more sense.

pkg/deployment/depresource/new_deployment.go
```diff
@@ -24,46 +24,45 @@ import (
...
+	Name                  string
+	Version               string
+	DeploymentTemplateID  string
+	Region                string
+	ApmEnable             bool
+	AppsearchEnable       bool
+	Writer                io.Writer
+	Plugins               []string
+	TopologyElements      []string
+	ElasticsearchInstance InstanceParams
```
marclop: Same comment here about the fields from ElasticsearchInstance to ElasticsearchResource

pkg/deployment/depresource/new_deployment.go
karencfv: I did at first and then it looked like `depresource.ResourceParams` a bit stuttery 😄

PR #254, opened 2020-04-07
ecctl: Prefix config parsing error

### LINKED ISSUE #220: Show a user-friendly error when config file doesn't exist or is not readable
## Readiness Checklist

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior
When the config file used does not exist, the user should be informed with a user-friendly message

## Current Behavior
No error is reported. Instead, a not-very-helpful message is displayed
```
$ ecl platform region list --config wrong-file
1 error occurred:
	* apikey or user and pass must be specified
```

## Possible Solution
Check for config file existence and show a user-friendly message

## Steps to Reproduce
Just run any command by providing a config file value pointing to a file that doesn't exist

## Context
Maybe this is happening also even if the config file exists but its contents cannot be read.
I assume we need to handle that case too

## Your Environment
* Version used: 1.0.0-beta2

### LINKED ISSUE #223: [UX] explicitly require ecctl initialization
In ecctl-beta2, if a command is run without ecctl having an init file, the following error message is received:
image

**Lets please add text that requires ecctl to be initialized.** 

Something along the lines of:
"apikey or user and pass must be specified using the ecctl init command"

marclop:
## Description

This change changes the `ecctl.<struct>.Validate()` funcitons
multierrors to use `cloud-sdk-go/pkg/multierror.Prefixed` to improve the
quality of our errors to always use prefixes.

Additionally, changes the behavior when the ecctl.App fails to be
initialized and the config file used is empty (unexisting, or failed to
read) to throw an error which instructs the user to run "ecctl init".

## Related Issues

Resolves #223.
Resolves #220.

## How Has This Been Tested?

```console
$ ecctl user
uninitialized ecctl config file, please use "ecctl init" to initialize it: 1 error occurred:
    * invalid configuration options specified: apikey or user and pass must be specified
```

## Types of Changes


cmd/root.go
```diff
@@ -78,7 +78,16 @@ var RootCmd = &cobra.Command{
 			return err
 		}
 
-		return initApp(cmd, defaultClient, defaultViper)
+		err := initApp(cmd, defaultClient, defaultViper)
+		// When no config file has been read and initApp returns an error, tell
+		// the user how to initialize the application.
+		if err != nil && defaultViper.ConfigFileUsed() == "" {
+			return multierror.NewPrefixed(
+				`missing ecctl config file, please use "ecctl init" to initialize it`, err,
```
Kushmaro, line 86: @marclop just a minor fix - "... please use "ecctl init" command, to initialize ecctl"

cmd/root_test.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
+	"github.com/elastic/cloud-sdk-go/pkg/multierror"
```
karencfv, line 32: this import block needs formatting

PR #259, opened 2020-04-08
Fix some typos

acrewdson:


pkg/deployment/depresource/appsearch_test.go
```diff
@@ -159,7 +159,7 @@ func TestNewAppSearch(t *testing.T) {
 				),
 				Region: "ece-region",
 			}},
-			err: errors.New("deployment: the an ID template is not configured for AppSearch. Please use another template if you wish to start AppSearch instances"),
+			err: errors.New("deployment: an ID template is not configured for App Search. Please use another template if you wish to start App Search instances"),
```
karencfv, line 162: Let's keep the "the" at the beginning of the sentence to get the tests passing again :) It's part of the error message.

I see how it can look like a grammatical error on this test, but that's because the [id](https://github.com/elastic/ecctl/pull/259/files#diff-0b76f5a649e4d46cd0bc712e82bd6079R84) that we pass in the mock response is "an ID". Normally the id will be something more along the lines of "data.default" or something like that. Which will make the real error message look like: "deployment: the data.default template is not configured..." 
```suggestion
			err: errors.New("deployment: the an ID template is not configured for App Search. Please use another template if you wish to start App Search instances"),
```

acrewdson: Thanks @karencfv for the explanation, makes sense 👍 I'll fix that.

karencfv:
One more thing, just noticed sorry. Looks like you need to sign your commit :)

Here are the instructions https://help.github.com/en/github/authenticating-to-github/about-commit-signature-verification

acrewdson:
@karencfv no problem, should be signed now 👍 thanks!

karencfv:
Thanks @acrewdson !

PR #287, opened 2020-05-12
release: Fix git config and changelog format

marclop:
## Description

Fixes the Changelog notes to have the version in the first line as it's
the title that GitHub will use for the release.

Additionally, fixes the issue where the git config doesn't have any user
set, now the script configures it when the settings are missing.

Last, corrects the `GITHUB_USER` to `elasticcloudmachine` when
updating the home-brew tap.

## Motivation and Context

Failing release action.

### DIFF
--- a/scripts/goreleaser-post-actions.sh
+++ b/scripts/goreleaser-post-actions.sh
@@ -17,6 +17,14 @@ for f in dist/*.{tar.gz,deb,rpm,txt}; do
 done
 
 # Create the actual Github Release.
+if [[ -z $(git config --get user.email) ]]; then
+    git config user.email "cloud-delivery@elastic.co"
+fi
+
+if [[ -z $(git config --get user.name) ]]; then
+    git config user.name "elasticcloudmachine"
+fi
+
 hub release create -F notes/${VERSION}.md ${VERSION}
 
 # Update the brew tap formula
--- a/scripts/goreleaser-post-actions.sh
+++ b/scripts/goreleaser-post-actions.sh
@@ -16,15 +16,18 @@ for f in dist/*.{tar.gz,deb,rpm,txt}; do
     ${f} s3://download.elasticsearch.org/downloads/ecctl/$(echo ${VERSION}| sed 's/^v//')/
 done
 
+# If in the CI environment and the user credentials aren't set, set them.
+if [[ ${CI} ]]; then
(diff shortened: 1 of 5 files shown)


scripts/goreleaser-post-actions.sh
```diff
@@ -17,6 +17,14 @@ for f in dist/*.{tar.gz,deb,rpm,txt}; do
 done
 
 # Create the actual Github Release.
+if [[ -z $(git config --get user.email) ]]; then
```
rhass, line 20: Why are we adding `git config` to this script?

marclop: I've pushed another commit: 614d377 and 4dc17a8, which only would set this if the environment variable `CI` is set to `true`. This will only happen within the GitHub action which is executing the release.

If you check the execution of the release here:
https://github.com/elastic/ecctl/runs/665265418?check_suite_focus=true, you'll see that the `git commit` fails because the git configuration isn't set.

The actual script which. runs that is in [`scripts/update-brew-tap.sh`](https://github.com/elastic/ecctl/blob/master/scripts/update-brew-tap.sh)

rhass: Rather than changing the configuration, you can use environment variables to accomplish the same thing. 

https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

We do this with the cloud cut-release jobs: 
https://github.com/elastic/cloud/blob/e71b84ab776c2638d63da83e1c3e96537a8d6902/.ci/jobs/cut-release-ece.yml#L151-L153

In out case however, [we leverage a Jenkins plugin](https://github.com/elastic/cloud/blob/e71b84ab776c2638d63da83e1c3e96537a8d6902/.ci/jobs/cut-release-ece.yml#L137) to set the values to the person who executed the cut release job so that we can have a more rich commit history.

Either way you do it -- you shouldn't need to modify the git configuration with the environment variables.

marclop: ah that's cool. I'll correct it then!

marclop: I addressed your comments in c580e75

marclop: Actually, I added them to the `env` object in the `yaml` specification for the release action (b37494c).

PR #296, opened 2020-05-28
Docs: Fix links to ECE API definitions (1.0.0-beta2)

kilfoyle:
We're doing some reorganization of the ECE book, so this fixes two links that will break as a result of that change. I believe we'll have to merge this at the same time as the ECE changes to avoid a docs build break.


kilfoyle:
Thanks Karen!  I had some difficulty with that `v1.0.0-beta2` release and wasn't able to access it locally in order to push to my fork.  But, Nik has helped me and we're no longer building the docs from that branch anyway, so I just closed those two PRs.  This PR is based off of my fork so I trust that's okay.

PR #297, opened 2020-05-28
Docs: Fix links to ECE API APM payload definition (v1.0.0-beta2)

kilfoyle:
We're doing some reorganization of the ECE book, so this fixes the APM link that will break as a result of that change. I believe we'll have to merge this at the same time as the ECE changes to avoid a docs build break.


kilfoyle:
Closing...  this branch is no longer being built.

PR #305, opened 2020-06-11
constructor: Populate constructorapi Region field

marclop:
## Description

Updates the constructor commands to use the region per API call.


karencfv:
merging as I need it for my PR :)

PR #330, opened 2020-06-29
cmd/util/testutils: Add RunCmdAssertion test function

marclop:
## Description

Adds a new function to the `testutils` package which allow easier
testing to be written on the `cmd` layer.

All the heavy lifting is done by `testutils.RunCmdAssertion` on the test
function, namely it:

    1. Initializes the `ecctl.App` singleton from a `testutils.MockCfg`.
    2. Sets the cobra `cmd.Command` Err and Out to `*Bytes.Buffer`s.
    3. Executes the command with the specified Args.
    4. Asserts the returned errors by `cmd.Execute()`
    5. Asserts the written data to both the `Err` and `Out` buffers.
    6. `ecctl.Cleanup` is called before returning so no instance is set.

It is important to note that global flags won't be parsed and instead,
must be provided as values to `testutils.MockCfg`. As new global flags
are added, the structure should be maintained to contain these.

Interaction with the `stdin` hasn't been written yet, but it will be a
follow-up patch.

Last, the tracker utils have been modified to use `DefaultTestFrequency`
when the host equals `api.DefaultMockHost` for ease of testing.

## Motivation and Context

Easier cmd layer testing!

## How Has This Been Tested?

`make unit`

## Types of Changes

- [x] Refactoring (improves code quality but has no user-facing effect)

### DIFF
--- a/cmd/util/testutils/cmd.go
+++ b/cmd/util/testutils/cmd.go
@@ -46,14 +46,14 @@ type Assertion struct {
 	Stderr string
 }
 
-// TestCommand tests a `*cobra.Command` and uses the testing.T struct to
+// RunCmdAssertion tests a `*cobra.Command` and uses the testing.T struct to
 // return any unmatched assertions.
-func TestCommand(t *testing.T, args Args, assertion Assertion) {
+func RunCmdAssertion(t *testing.T, args Args, assertion Assertion) {
 	cfg := fillDefaults(args.Cfg)
 	cfg.Out = new(bytes.Buffer)
 	cfg.Err = new(bytes.Buffer)
 
-	defer MockApp(t, cfg)()
+	defer mockApp(t, cfg)()
 
 	// Set the command arguments.
 	args.Cmd.Root().SetArgs(args.Args)
@@ -66,8 +66,16 @@ func TestCommand(t *testing.T, args Args, assertion Assertion) {
 	args.Cmd.SetOutput(cfg.Out)
 	args.Cmd.SetErr(cfg.Err)
 
-	if err := args.Cmd.Execute(); !assert.Equal(t, assertion.Err, err) {
-		t.Error(err)
+	err := args.Cmd.Execute()
+
+	assertCmd(t, args, assertion, err,
+		strings.Contains(strings.Join(args.Args, " "), "--track"),
+	)
+}
+
+func assertCmd(t *testing.T, args Args, want Assertion, execErr error, track bool) {
+	if !assert.Equal(t, want.Err, execErr) {
+		t.Error(execErr)
 	}
 
 	if buf, ok := args.Cmd.OutOrStdout().(*bytes.Buffer); ok {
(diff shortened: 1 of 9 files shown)


marclop:
Commit breakdown:

* c5b20de: Adds the framework as described in the PR.
* 1386562: Adds tests for the `deployment create` command, with and without tracking.

cmd/util/testutils/cmd.go
```diff
@@ -0,0 +1,94 @@
...
+		if got != assertion.Stdout {
+			t.Errorf(`"Got stdout "%s" != want "%s"`, got, assertion.Stdout)
+		}
+	}
+
+	if buf, ok := args.Cmd.ErrOrStderr().(*bytes.Buffer); ok {
+		if got := buf.String(); got != assertion.Stderr {
+			t.Errorf(`"Got stderr "%s" != want "%s"`, got, assertion.Stderr)
+		}
+	}
```
karencfv, line 93: I think I'd like this assertion code outside of the `TestCommand()` function, and in it's own function. What do you think?

cmd/util/testutils/cmd.go
```diff
@@ -0,0 +1,94 @@
...
+// Assertion to use for tests.
+type Assertion struct {
+	Err    error
+	Stdout string
+	Stderr string
+}
+
+// TestCommand tests a `*cobra.Command` and uses the testing.T struct to
+// return any unmatched assertions.
+func TestCommand(t *testing.T, args Args, assertion Assertion) {
```
karencfv, line 51: I'm not sure I like `testutils.TestCommand` 🤔 How about `cmdutils.Test` or `testutils.AssertCommand` ? or something better that you might come up with 😄

cmd/util/testutils/testutils.go
```diff
@@ -18,25 +18,78 @@
...
+		Format:       cfg.Format,
+		Host:         fmt.Sprintf("https://%s", api.DefaultMockHost),
+		APIKey:       defaultAPIKey,
+		Force:        cfg.Force,
+		Verbose:      cfg.Verbose,
+	}
+}
+
+// MockApp initiates a mocked app from a MockCfg.
+func MockApp(t *testing.T, cfg MockCfg) func() {
```
karencfv, line 89: This doesn't need to be exported any more, does it?

cmd/util/testutils/cmd.go
marclop: you mean as a private function?

cmd/util/testutils/cmd.go
marclop: haha fair point, maybe `testutils.RunAssertion`, or just `Run`? `cmdutils.TestCommand` sounds good too.

cmd/util/testutils/testutils.go
marclop: Not really no, but it'd allow us to call it externally if we need? although I'm not sure we need to anymore

cmd/util/testutils/cmd.go
karencfv: yep :)

cmd/util/testutils/cmd.go
karencfv: I think just `Run` might be a bit confusing if the name of the package is `testutils`. I think I like `testutils.RunCmdAssertion` or `cmdutils.RunAssertion`. What do you think?

cmd/util/testutils/testutils.go
karencfv: I don't think we need to any more 🤔 If we really need it we can export it later down the track. I also wouldn't want someone to find this function and skip the `TestCommand` function or whatever. What do you think?

marclop: yeah Ok, I'll hide this function

marclop:
@karencfv I've addressed most of your comments, but left the two definitions.

I think it's better to have two deployment definitions with different values than 1, it doesn't mean that we should always use 2 different "provider" resource definitions but I did add them to the test to have different data. I've added the APM resource to the AWS definition.

PR #336, opened 2020-07-07
cmd: fix testutil.newConfig ignored region

marclop:
## Description

Fixes a bug on the testutils which caused the `Region` property to not
be populated with either the default value or the overriding value of a
`MockCfg`.


karencfv:
Merged this myself as I need it for my PR :)

Issue #334, opened 2020-07-06
Reinstate flag based deployment create command


karencfv:
## Description

The [work done](https://github.com/elastic/ecctl/pull/224) for a flag based `ecctl deployment create` command should be reinstated.

Additionally, some refactoring needs to be done to use the new deployment template endpoints instead of the ones it's currently calling.

PR #239, opened 2020-04-02
cmd: Temporarily remove ability to create a flag based deployment

karencfv:
## Description
This change temporarily removes the ability to create a flag based deployment until reads to deployment templates are available to all users.

## Motivation and Context
Currently only ECE users have permissions to perform reads on deployment templates. To create a flag based deployment all users would need these permissions. We will wait until these endpoints are available to all to reinstate flag based deployment creation.

## How Has This Been Tested?
Manually by running the command and by running `make docs`

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (improves code quality but has no user-facing effect)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation

## Readiness Checklist

- [x] My code follows the code style of this project
- [x] My change requires a change to the documentation
- [x] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #341, opened 2020-07-14
deployment: create flags reenabled and Enterprise Search support

karencfv:
**Requires https://github.com/elastic/cloud-sdk-go/pull/168 to be merged**

## Description
Re-enables `ecctl deployment create` flags (https://github.com/elastic/ecctl/pull/239)

Also, there is now support for Enterprise Search and the default deployment 
template is chosen based on cloud provider.

## Related Issues
Closes https://github.com/elastic/ecctl/issues/334

## How Has This Been Tested?
tests and manually running the command many times

## Types of Changes

## Readiness Checklist

### DIFF
--- a/cmd/deployment/create.go
+++ b/cmd/deployment/create.go
@@ -33,13 +33,13 @@ import (
 )
 
 var createCmd = &cobra.Command{
-	Use:     "create {--file | --size <int> --zones <string> | --topology-element <obj>}",
+	Use:     "create {--file | --es-size <int> --es-zones <int> | --topology-element <obj>}",
 	Short:   "Creates a deployment",
 	PreRunE: cobra.NoArgs,
-	// Switch back to non-temp constants when reads for deployment templates are available on ESS
-	Long:    createLongTemp,
-	Example: createExampleTemp,
+	Long:    createLong,
+	Example: createExample,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		var file, _ = cmd.Flags().GetString("file")
 		var track, _ = cmd.Flags().GetBool("track")
 		var generatePayload, _ = cmd.Flags().GetBool("generate-payload")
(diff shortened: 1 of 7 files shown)


cmd/deployment/create.go
```diff
@@ -190,25 +196,32 @@ func init() {
...
+		dt = "azure-io-optimized"
+	case "gcp":
+		dt = "gcp-io-optimized"
+	case "ece-region":
+		dt = "default"
+	default:
+		dt = "aws-io-optimized-v2"
+	}
+
+	return dt
```
marclop, line 226: I would have used early returns since it seems more declarative than assigning vars, but it doesn't really matter.

```suggestion
	switch region {
	case "azure":
		return "azure-io-optimized"
	case "gcp":
		return "gcp-io-optimized"
	case "ece-region":
		return "default"
	default:
		return "aws-io-optimized-v2"
	}
```

docs/ecctl_deployment_create.adoc
```diff
@@ -6,38 +6,90 @@ Creates a deployment
...
 
-Creates a deployment which is defined from a file definition using the --file=+++<file path="">+++(shorthand: -f) flag.+++</file>+++
+Creates a deployment which can be defined through flags or from a file definition.
+Sane default values are provided, making the command work out of the box even when no parameters are set.
+When version is not specified, the latest available stack version will automatically be used.
+These are the available options:
 
-You can create a definition by using the sample JSON:
-  https://elastic.co/guide/en/cloud/current/ec-api-deployment-crud.html#ec_create_a_deployment
+* Simplified flags to set size and zone count for each instance type (Elasticsearch, Kibana, APM and App Search).
```
marclop, line 14: I can't comment on line :27 of `create_help.go`, but I think we need to add the Enterprise Search too, here.

Issue #335, opened 2020-07-06
ecctl GA


karencfv:
## Description
ecctl will be GA soon 🎉 This is a list to track the tasks we need to finish.

### Done

- [x] Move all APIs from ecctl to cloud-sdk-go https://github.com/elastic/ecctl/issues/275
- [x] Refactor platform APIs to require `Region` parameter https://github.com/elastic/cloud-sdk-go/issues/129
- [x] Change `ecctl platform stack` command and subcommands to `ecctl stack` for parity with the API https://github.com/elastic/ecctl/pull/329
- [x] Ability for deployment template APIs to use the `deployment` format. https://github.com/elastic/cloud-sdk-go/issues/159 https://github.com/elastic/ecctl/pull/332
- [x] Update client/models to MS39 and update `user update` API https://github.com/elastic/cloud-sdk-go/issues/160
- [x] Update client/models to MS41 https://github.com/elastic/cloud-sdk-go/pull/167 @karencfv 
- [x] Flag based deployment create https://github.com/elastic/ecctl/issues/334 @karencfv #341 (The PR has been merged but the issue about deployment template ID availability needs to be solved) 
  - [x] Resource Get @karencfv https://github.com/elastic/cloud-sdk-go/pull/170
- [x] Add custom ecctl headers (https://github.com/elastic/ecctl/issues/333) @marclop
- [x] Make sure Enterprise Search is supported in all commands it should be.
  - [x] Vacate
  - [x] flag-based Deployment create https://github.com/elastic/ecctl/pull/341 https://github.com/elastic/cloud-sdk-go/pull/168 
  - [x] Various mentions of resourcekinds @karencfv https://github.com/elastic/ecctl/pull/342 https://github.com/elastic/cloud-sdk-go/pull/170
- [x] Use deployment templates new APIs instead of the old platform API ones. Also replace old `platform deployment-template` command and subcommands or mark them as deprecated @marclop elastic/cloud-sdk-go#171 and #343.
- [x] Make sure our documentation in ecctl and cloud-sdk-go is up to date with recent changes @karencfv https://github.com/elastic/ecctl/pull/347
- [x] Add `--generate-update-payload` to `ecctl deployment show` command (#355 and elastic/cloud-sdk-go#181)
- [x] Move the `deptemplateapi.Get` API calls to `deptemplateapi.List` with the instance filter for `depresource.NewPayload` (cloud-sdk-go). @marclop (elastic/cloud-sdk-go#178) (#348)
[...]

PR #343, opened 2020-07-16
deployment: Add template v2 and remove platform commands

marclop:
## Description

Removes the platform `deployment-template` commands.
Adds a new `deployment template` command and all its CRUD subcommands:

    * `deployment template create`
    * `deployment template list`
    * `deployment template show`
    * `deployment template delete`
    * `deployment template update`

## Review comments

I've split the commits so that the reviewing process is easier.

## Related Issues

#335 

## Types of Changes


cmd/deployment/template/show.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+
+		return ecctl.Get().Formatter.Format("", res)
+	},
+}
+
+func init() {
+	Command.AddCommand(showCmd)
+	showCmd.Flags().Bool("hide-instance-configurations", false, "Hides instance configurations - only visible when using the JSON output")
+	showCmd.Flags().String("stack-version", "", "It will cause the returned deployment templates which are valid for the specified stacks version.")
```
karencfv, line 53: is this flag necessary if you're only returning the specified template?

```suggestion
	showCmd.Flags().String("stack-version", "", "Optional filter to only return deployment templates which are valid for the specified stack version.")
```

cmd/deployment/template/show.go
```diff
@@ -0,0 +1,54 @@
...
+	Short:   "Displays a deployment template",
+	PreRunE: cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		stackVersion, _ := cmd.Flags().GetString("stack-version")
+		hideIC, _ := cmd.Flags().GetBool("hide-instance-configurations")
+
+		res, err := deptemplateapi.Get(deptemplateapi.GetParams{
+			API:                        ecctl.Get().API,
+			Region:                     ecctl.Get().Config.Region,
+			StackVersion:               stackVersion,
```
karencfv, line 38: is this necessary if you're only returning the specified template?

cmd/deployment/template/update.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/elastic/cloud-sdk-go/pkg/models"
+	sdkcmdutil "github.com/elastic/cloud-sdk-go/pkg/util/cmdutil"
+	"github.com/spf13/cobra"
+
+	"github.com/elastic/ecctl/pkg/ecctl"
+)
+
+var updateCmd = &cobra.Command{
+	Use:     "update",
+	Short:   "Updates an existing deployment template",
```
karencfv, line 35:
```suggestion
	Short:   cmdutil.AdminReqDescription("Updates an existing deployment template"),
```

cmd/deployment/template/update.go
```diff
@@ -0,0 +1,63 @@
...
+			TemplateID: templateID,
+			Request:    req,
+		})
+	},
+}
+
+func init() {
+	Command.AddCommand(updateCmd)
+	updateCmd.Flags().Bool("hide-instance-configurations", false, "Hides instance configurations - only visible when using the JSON output.")
+	updateCmd.Flags().String("file", "", "It will cause the returned deployment templates which are valid for the specified stacks version.")
```
karencfv, line 59:
```suggestion
	updateCmd.Flags().StringP("file", "f", "", "It will cause the returned deployment templates which are valid for the specified stacks version.")
```

cmd/deployment/template/update.go
```diff
@@ -0,0 +1,63 @@
...
+	updateCmd.Flags().String("template-id", "", "It will create the deployment template with the specified ID rather than auto-generating an ID.")
```
karencfv, line 60: Should the deployment template ID be passed as an argument (like `show`) instead of a flag given it's not an optional parameter like in `create`?

cmd/platform/deployment-template/command.go
```diff
@@ -39,8 +39,10 @@ const (
 
 // Command represents the top level deployment-template command.
 var Command = &cobra.Command{
-	Use:     "deployment-template",
-	Short:   cmdutil.AdminReqDescription("Manages deployment templates"),
+	Use: "deployment-template",
```
karencfv, line 42: I am tempted into removing these completely and moving them over just like we did with the `clusters` API ones 🤔 what do you think. Last chance to remove things easily before GA!

cmd/deployment/template/show.go
marclop: Apparently it's a valid argument, so I left it because it's available on the API. I guess it might be useful if you're trying to verify that a DT is compatible with stack version X.Y.Z

cmd/deployment/template/show.go
marclop: Optional filter to only return deployment templates which are valid for the specified stack version.

cmd/deployment/template/update.go
marclop: Yeah I thought about this and my logic was to keep the API closely with the `create` operation since that one has a `--template-id` optional argument.

karencfv: Should `show` have the `--template-id` flag then? It feels odd to have one command take the template ID as an argument, and another as a flag parameter 🤔 

I think the flag's description should change too, this description is for `create` 😄

cmd/deployment/template/show.go
karencfv: true 🤔

PR #344, opened 2020-07-16
deployment: Add elasticsearch keystore commands

marclop:
## Description

Adds a new `elasticsearch` subcommand to the `deployment` command and a
`keystore` subcommand to the `deployment elasticsearch` command. Results
in:

* `deployment elasticsearch keystore show`
* `deployment elasticsearch keystore update|set`

## Related Issues

Resolves #213 

## How Has This Been Tested?

Unit and manually tested against an ESS deployment.

## Types of Changes

- [x] New feature (non-breaking change which adds functionality)

PR #346, opened 2020-07-20
allocator: Fix Array parsing to slice parsing

marclop:
## Description

Fixes a bug where the `--target` parameter was being ignored due to the
flag parsing using `GetStringSlice` when the flag decaration is `Array`.

The bug changes the type from an ArrayString to SliceString.

Additionally, fixes `--maintenance` call which is missing a region value.

## Motivation and Context

`--target` flag was being incorrectly parsed.

## How Has This Been Tested?

Manually.

## Types of Changes


marclop:
unit tests are failing due to broken docs and are fixed by #344

PR #348, opened 2020-07-21
deployment: Fix create command against ESS

marclop:
## Description

Fixes the `deployment create` command to be able to pull the deployment
templates by listing them using the API.

Additionally fixes the missing version on statless resources when the
version flag is not specified and it's a flag-based create.

Last, fixes the create tests which weren't testing the create workflow
on the flag-based scenario.

## Related Issues

Requires elastic/cloud-sdk-go#178 to be merged.
Part of #335

### DIFF
--- a/cmd/deployment/create.go
+++ b/cmd/deployment/create.go
@@ -89,20 +89,22 @@ var createCmd = &cobra.Command{
 			dt = setDefaultTemplate(region)
 		}
 
+		dtAsList, _ := cmd.Flags().GetBool("dt-as-list")
 		if payload == nil {
 			var err error
 			payload, err = depresourceapi.NewPayload(depresourceapi.NewPayloadParams{
-				API:                    ecctl.Get().API,
-				Name:                   name,
-				DeploymentTemplateID:   dt,
-				Version:                version,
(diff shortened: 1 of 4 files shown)


cmd/deployment/create.go
```diff
@@ -200,6 +206,10 @@ func init() {
 	createCmd.Flags().String("enterprise-search-ref-id", "main-enterprise_search", "Optional RefId for the Enterprise Search deployment")
 	createCmd.Flags().Int32("enterprise-search-zones", 1, "Number of zones the Enterprise Search instances will span")
 	createCmd.Flags().Int32("enterprise-search-size", 4096, "Memory (RAM) in MB that each of the Enterprise Search instances will have")
+
+	// Remove in the next version.
+	createCmd.Flags().Bool("dt-as-list", true, "")
```
karencfv, line 211: In this case, instead of having a flag and hiding it. might it not make it simpler to assign `true` directly to `DeploymentTemplateAsList`?

marclop: I did this so that we could keep the normal (Get) flow in the tests and toggle the `List()` with the flag.

PR #352, opened 2020-07-22
docs: specify all ECE specific commands and add icons to .adoc files

karencfv:
## Description
This PR is separated in to two.

The first commit adds ECE specific text to all subcommands not just commands as it was before.

The second commit substitutes this text for `{ece-icon}` so our online documentation looks better with an icon instead of text.

@nrichers I'm assuming this will work fine? I looked at the kibana repo and didn't find any actual icon added to their repo

## Related Issues
Related: https://github.com/elastic/ecctl/issues/335

## Motivation and Context
Make the docs pretty 😊 

## Types of Changes

- [x] Documentation

### DIFF
--- a/cmd/auth/command.go
+++ b/cmd/auth/command.go
@@ -26,7 +26,7 @@ import (
 // Command is the auth subcommand
 var Command = &cobra.Command{
 	Use:     "auth",
-	Short:   "Manages authentication",
+	Short:   "Manages authentication settings",
 	PreRunE: cobra.MaximumNArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		cmd.Help()
--- a/cmd/auth/command.go
+++ b/cmd/auth/command.go
@@ -26,7 +26,7 @@ import (
 // Command is the auth subcommand
 var Command = &cobra.Command{
 	Use:     "auth",
-	Short:   "Manages the platform auth",
+	Short:   "Manages authentication",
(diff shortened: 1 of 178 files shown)


cmd/auth/command.go
```diff
@@ -26,7 +26,7 @@ import (
 // Command is the auth subcommand
 var Command = &cobra.Command{
 	Use:     "auth",
-	Short:   "Manages the platform auth",
+	Short:   "Manages authentication",
```
marclop, line 29: Maybe add `settings` at the end?
```suggestion
	Short:   "Manages authentication settings",
```

scripts/generate-docs.sh
```diff
@@ -39,6 +39,9 @@ done
 # Strip extraneous .adoc out of section IDs and xrefs
 sed -i'.bak' -e 's/\.adoc//g' ecctl*.adoc
 
+# Replace ECE availability text for ece icon
+sed -i'.bak' -e 's/(Available for ECE only)/{ece-icon}/g' ecctl*.adoc
```
nrichers, line 43: Kudos on doing this for the Asciidoc even through the Markdown doesn't really support it. ❤️

karencfv: 🙇‍♀️

karencfv:
@nrichers yeah, I think we could do that 🤔 unless @marclop has any objection :)

PR #355, opened 2020-07-27
deployment: Add --generate-update-payload to show

marclop:
## Description

Adds a new `--generate-update-payload` flag to the `ecctl deployment show`,
cmd updates the help for both `show` and `update`  commands and updates
the module `elastic/cloud-sdk-go` to the latest commit.

Adds unit tests for the `--generate-update -payload` flag on the show cmd.

## Related Issues

Part of #335 

## Types of Changes


cmd/deployment/show.go
```diff
@@ -32,7 +35,10 @@ const showExample = `
   ecctl deployment show <deployment-id> --kind kibana
 
 * Shows apm resource information from a given deployment with a specified ref-id.
-  ecctl deployment show <deployment-id> --kind apm --ref-id apm-server`
+  ecctl deployment show <deployment-id> --kind apm --ref-id apm-server
+
+* Return the current deployment state as a valid create / update payload.
```
karencfv, line 40:
```suggestion
* Return the current deployment state as a valid update payload.
```

I don't think this would work for create, would it? It has a `prune_orphans` field

cmd/deployment/update_help.go
```diff
@@ -0,0 +1,116 @@
...
+// under the License.
+
+package cmddeployment
+
+const updateLong = `updates a deployment from a file definition, defaulting prune_orphans=false, making the default
+update action safe for partial updates, to override this behavior toggle --prune-orphans.
+To track the changes toggle the --track flag.
+
+It is possible to use "--generate-payload" as: "ecctl deployment show --generate-payload ..."
+to obtain a valid create / update payload from an existing deployment which can be manually edited in
```
karencfv, line 25:
```suggestion
to obtain a valid update payload from an existing deployment which can be manually edited in
```
same as above, can we double check?

cmd/deployment/update_help.go
```diff
@@ -0,0 +1,116 @@
...
+    },
+    {
+      "elasticsearch_cluster_ref_id": "my-es-cluster",
+      "id": "90c3c9566f454861b0dc935c5c7420d8",
+      "kind": "apm",
+      "ref_id": "my-apm-instance",
+      "region": "ece-region"
+    }
+  ]
+}
```
karencfv, line 91: I remember from a review @nrichers gave me, he mentioned something about it not being necessary to write out the output in examples. I think the idea is to keep it simple, that's why I removed a lot of text from the create command. Please correct me if I'm wrong Nik! 🙏

cmd/deployment/update_help.go
```diff
@@ -0,0 +1,116 @@
...
+    "apm": [
+      "90c3c9566f454861b0dc935c5c7420d8"
+    ],
+    "appsearch": [],
+    "elasticsearch": [],
+    "kibana": [
+      "3617594f01074b76a5ca4f903f9d33ec"
+    ]
+  }
+}`[1:]
```
karencfv, line 116: same as above

cmd/deployment/show.go
marclop: That's right!

cmd/deployment/show.go
```diff
@@ -90,4 +111,5 @@ func initShowFlags() {
 	showCmd.Flags().Bool("plan-history", false, "Shows the deployment plan history")
 	showCmd.Flags().BoolP("metadata", "m", false, "Shows the deployment metadata")
 	showCmd.Flags().BoolP("settings", "s", false, "Shows the deployment settings")
+	showCmd.Flags().Bool("generate-payload", false, "Outputs a JSON formatted payload which can be used as an argument for the --file flag on update command.")
```
nrichers, line 114:
```suggestion
	showCmd.Flags().Bool("generate-payload", false, "Outputs JSON which can be used as an argument for the --file flag with the update command.")
```

(If this is technically accurate, can you simplify the wording as suggested?)

cmd/deployment/update_help.go
```diff
@@ -0,0 +1,70 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package cmddeployment
+
+const updateLong = `updates a deployment from a file definition, defaulting prune_orphans=false, making the default
```
nrichers, line 20:
```suggestion
const updateLong = `updates a deployment from a file definition. Defaults prune_orphans=false, making the default
```

(There is a long sentence, so I suggested breaking it up with this comment and the next.)

cmd/deployment/update_help.go
```diff
@@ -0,0 +1,70 @@
...
+update action safe for partial updates, to override this behavior toggle --prune-orphans.
```
nrichers, line 21:
```suggestion
update action safe for partial updates. To override this behavior, toggle --prune-orphans.
```

cmd/deployment/show.go
```diff
@@ -32,7 +35,10 @@ const showExample = `
   ecctl deployment show <deployment-id> --kind kibana
 
 * Shows apm resource information from a given deployment with a specified ref-id.
-  ecctl deployment show <deployment-id> --kind apm --ref-id apm-server`
+  ecctl deployment show <deployment-id> --kind apm --ref-id apm-server
```
nrichers, line 38: I mentioned this to Karen already, but the style guide at Elastic says to use `UPPERCASE_VARIABLE_NAMES_WITH_UNDERSCORES`, sometimes prefixed with `$`. I wouldn't change this here if `<variable-name>` is what you use in other places already, but it does put your docs out of sync with the ECE API reference for example (but even there we aren't always consistent).

cmd/deployment/show.go
```diff
@@ -32,7 +35,10 @@ const showExample = `
...
+
+* Return the current deployment state as a valid update payload.
+  ecctl deployment show <deployment id> --generate-payload > update.json`
```
nrichers, line 41: Does this line need a leading "`"?

cmd/deployment/show.go
marclop: Ah, then we should fix it in the upcoming versions.

cmd/deployment/show.go
marclop: ah no, the enclosing with backpacks in Golang means a string literal with line breaks allowed in there, basically it allows us to define multi-line literals in a var.

cmd/deployment/show.go
```diff
@@ -111,5 +111,5 @@ func initShowFlags() {
 	showCmd.Flags().Bool("plan-history", false, "Shows the deployment plan history")
 	showCmd.Flags().BoolP("metadata", "m", false, "Shows the deployment metadata")
 	showCmd.Flags().BoolP("settings", "s", false, "Shows the deployment settings")
-	showCmd.Flags().Bool("generate-payload", false, "Outputs a JSON formatted payload which can be used as an argument for the --file flag on create / update commands.")
+	showCmd.Flags().Bool("generate-payload", false, "Outputs a JSON formatted payload which can be used as an argument for the --file flag on update command.")
```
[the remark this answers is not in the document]
marclop: hm... I think that might be more accurate, yes

PR #356, opened 2020-07-27
deployment: Change of terminology for create command

karencfv:
## Description
The `Name` field in the elasticsearch topology element has changed to `NodeType`
This patch updates the documentation to take that change into account, as well as
changing the `--topology-element` flag to `--es-node-topology` as an UX improvement.

## Related Issues
https://github.com/elastic/cloud-sdk-go/pull/182
https://github.com/elastic/ecctl/issues/335

## Motivation and Context
UX

## How Has This Been Tested?
by running the command and unit test

## Types of Changes

## Readiness Checklist

### DIFF
--- a/cmd/deployment/create.go
+++ b/cmd/deployment/create.go
@@ -34,7 +34,7 @@ import (
 )
 
 var createCmd = &cobra.Command{
-	Use:     "create {--file | --es-size <int> --es-zones <int> | --topology-element <obj>}",
+	Use:     "create {--file | --es-size <int> --es-zones <int> | --es-node-topology <obj>}",
 	Short:   "Creates a deployment",
 	PreRunE: cobra.NoArgs,
 	Long:    createLong,
@@ -51,7 +51,7 @@ var createCmd = &cobra.Command{
 		var esZoneCount, _ = cmd.Flags().GetInt32("es-zones")
 		var esSize, _ = cmd.Flags().GetInt32("es-size")
 		var esRefID, _ = cmd.Flags().GetString("es-ref-id")
-		var te, _ = cmd.Flags().GetStringArray("topology-element")
+		var topologyElements, _ = cmd.Flags().GetStringArray("es-node-topology")
 		var plugin, _ = cmd.Flags().GetStringSlice("plugin")
 
 		var kibanaZoneCount, _ = cmd.Flags().GetInt32("kibana-zones")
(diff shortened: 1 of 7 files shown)


cmd/deployment/create.go
```diff
@@ -185,7 +185,7 @@ func initFlags() {
 	createCmd.Flags().String("es-ref-id", "main-elasticsearch", "Optional RefId for the Elasticsearch deployment")
 	createCmd.Flags().Int32("es-zones", 1, "Number of zones the Elasticsearch instances will span")
 	createCmd.Flags().Int32("es-size", 4096, "Memory (RAM) in MB that each of the Elasticsearch instances will have")
-	createCmd.Flags().StringArrayP("topology-element", "e", nil, "Optional Elasticsearch topology element definition. See help for more information")
+	createCmd.Flags().StringArrayP("es-node-topology", "e", nil, "Optional Elasticsearch node topology element definition. See help for more information")
```
marclop, line 188: most of the flags are set as `StringSlice` not `StringArray`. I think we should change this to be `StringSliceP`. I know it's not related to your change, but I'm just seeing it now.

cmd/deployment/create.go
```diff
@@ -51,7 +51,7 @@ var createCmd = &cobra.Command{
 		var esZoneCount, _ = cmd.Flags().GetInt32("es-zones")
 		var esSize, _ = cmd.Flags().GetInt32("es-size")
 		var esRefID, _ = cmd.Flags().GetString("es-ref-id")
-		var te, _ = cmd.Flags().GetStringArray("topology-element")
+		var topologyElements, _ = cmd.Flags().GetStringArray("es-node-topology")
```
marclop, line 54: This should be changed to `GetStringSlice`

cmd/deployment/create.go
karencfv: The command doesn't work with `StringSlice` unfortunately :( I tried the command changing both lines you mentioned and I get `invalid argument "{\"size\": 1024, \"zone_count\": 2, \"node_type\": \"data\"}" for "-e, --es-node-topology" flag: parse error on line 1, column 1: bare " in non-quoted-field`

I think it's best to leave them as is, otherwise I feel like I might have to change several things. What do you think?

marclop: Ah that's right, that's why it's an array. sorry for the noise!

PR #360, opened 2020-08-05
cmd: Fix deployment note commands

karencfv:
## Description
Fixes an issue where the `ecctl deployment note` commands did not have the
`Region` parameter set.

Additionally, a small bug has been fixed where the ID that is meant to be passed to
the API is the Elasticsearch ID, not the deployment ID.

## How Has This Been Tested?
Manually running all the commands against an ECE environment.

## Types of Changes

### DIFF
--- a/cmd/deployment/note/command.go
+++ b/cmd/deployment/note/command.go
@@ -161,9 +161,9 @@ func init() {
 // `deployments/DEPLOYMENT_ID/notes` endpoint is not specific to elasticsearch.
 // For the time being the command layer will fetch the ES ID until the endpoint
 // accepts deployment IDs.
-func getElasticsearchID(deploymentID string, api *api.API) (string, error) {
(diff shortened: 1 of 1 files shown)


cmd/deployment/note/command.go
```diff
@@ -130,3 +156,19 @@ func init() {
...
+func getElasticsearchID(deploymentID string, ecAPI *api.API) (string, error) {
+	esID, err := deploymentapi.GetElasticsearchID(deploymentapi.GetParams{
+		API:          ecAPI,
+		DeploymentID: deploymentID,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	return esID, nil
```
marclop, line 173:
```suggestion
	return deploymentapi.GetElasticsearchID(deploymentapi.GetParams{
		API:          ecAPI,
		DeploymentID: deploymentID,
	})
```

karencfv: omg 🤦‍♀️

PR #363, opened 2020-08-13
cmd: Rename apikey to api-key and EC_API_KEY

marclop:
## Description

In order to align the environment variables to match the existing cloud
documentation, this patch renames the `--apikey` flag to `--api-key`,
which also makes the `EC_APIKEY` be changed to the desired `EC_API_KEY`
environment variable.

## Related Issues

elastic/terraform-provider-ec#39

## Motivation and Context

Aligning the config settings across products.

## Types of Changes


karencfv:
btw, I think the tile of the PR is wrong 😄

marclop:
@karencfv all the cli flags use a `-` to separate words, while env, yaml and json config use `_` as the key separators

PR #366, opened 2020-08-19
release: Fix git commit by adding user and email

marclop:
## Description

Configures the git username and email through `git config` command since
the environment variables seem to have no effect on the release target.

## Motivation and Context

Broken release GitHub action.

### DIFF
--- a/scripts/update-brew-tap.sh
+++ b/scripts/update-brew-tap.sh
@@ -29,7 +29,8 @@ cd /tmp/homebrew-tap
 hub fork --no-remote
 hub remote add fork https://github.com/${GITHUB_USER}/homebrew-tap
 hub add Formula/ecctl.rb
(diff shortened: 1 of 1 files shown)


scripts/update-brew-tap.sh
```diff
@@ -29,7 +29,8 @@ cd /tmp/homebrew-tap
 hub fork --no-remote
 hub remote add fork https://github.com/${GITHUB_USER}/homebrew-tap
 hub add Formula/ecctl.rb
-hub checkout -b f/update-ecctl-formula-to-${VERSION}
```
karencfv, line 32: why was this line removed?

marclop: Good catch, it shouldn't have

karencfv (approved):
Looks good!

PR #370, opened 2020-08-19
docs: Remove duplicated synopsis from doc files

### LINKED ISSUE #351: docs: Synopsis and Short description duplicated in generated .md files
## Description
When no `Long` description is provided in the command, the documentation 
generator duplicates the `Short` description and prints it twice.

Screen Shot 2020-07-22 at 7 10 33 PM

marclop:
## Description

Copies the markdown generator from github.com/spf13/cobra/doc, modifying
the source code so that it only writes down the `synopsis` header when
the `cmd.Long` is not empty.

Since it's impossible to use any of the existing function signatures to
generate the Markdown docs, copying the source and modifying it locally
is necessary.

## Related Issues

Resolves #351 

## How Has This Been Tested?

By running `make docs`

## Types of Changes

- [x] Documentation


marclop:
Thanks @nrichers, unfortunately this parts ways with how the markdown docs are generated, so we're forking that doc generating flow to our own repo.

I'll open an issue on the `spf13/cobra` repo to see if they'd like to include it upstream, but my feeling is not.

PR #378, opened 2020-09-10
allocator: Amend --all flag description

marclop:
## Description

Fixes the wording on the `platform allocator list --all` flag to be more
accurate with what's happening behind the scenes.

## Motivation and Context

Expectation that using the `--all` flag would return all the allocators in a region, but it's not the
flag's effect.

## Types of Changes

- [x] Documentation


karencfv (approved):
👌

Issue #376, opened 2020-09-08
bug: ecctl ignores EC_API_KEY environment variable


yokawasa:
## Question

ecctl version: 1.0.0

Q: How to skip ecctl init in executing ecctl command.  

My understanding is that I need to configure ecctl with `ecctl init` or set  environment variables like EC_API_KEY or 
 EC_USER & EC_PASS before I execute ecctl command like: 
`ecctl deployment show DEPLOYMENT_ID --kind elasticsearch --plans --host https://api.elastic-cloud.com`

What I've observed is that  even if I've set EC_API_KEY enviroment variable, i still come up with the following message in running `ecctl deployment show` like above.  Is this an expected behavior?  What I'd like to achieve is to run the ecctl command without configuring it with `ecctl init`  but with setting some enviroment variables like `EC_API_KEY`, etc. 

```
missing ecctl config file, please use the "ecctl init" command to initialize ecctl: 1 error occurred:
        * invalid configuration options specified: api_key or user and pass must be specified
```

NOTE:  in ecctl-1.0.0-beta3, I am able to run the command above as long as I set `EC_APIKEY` (NOT EC_API_KEY) environment variable.

marclop:
@yokawasa thanks for raising it. It seems that you have discovered a bug that's present on the released `v1.0.0` version. I've identified the bug and it'll be fixed in the upcoming `v1.0.1` version.

marclop:
Resolved by #381 . It will be available in the upcoming `v1.0.1` release

yokawasa:
@marclop  thank you for the prompt resultion!

PR #379, opened 2020-09-14
fix: ignored EC_API_KEY environment variable

marclop:
## Description

Fixes the problem where the `EC_API_KEY` environment variable wasn't
read.

The problem was that `viper` (correctly) unpacked the `EC_API_KEY` as
`api-key`, and since the `mapstructure` tags (in ecctl.Config) reference
`api_key`, the key is correctly un-marshaled when a config file is read,
but not when set via environment variables.

The fix is to set an alias `api_key` to the existing `api-key` key,
which is done in `cmd/root.go:143`.

    v.RegisterAlias("api_key", "api-key")

The rest of the patch is necessary to provide at least some unit tests
to be able to assert this behavior.

## Related Issues

Part of #376 

## How Has This Been Tested?

unit tested and tried against production ESS.

## Types of Changes

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -75,7 +75,7 @@ var RootCmd = &cobra.Command{
 	DisableAutoGenTag:      true,
 	BashCompletionFunction: bashCompletionFunc,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-		initConfig()
+		setupViper(defaultViper)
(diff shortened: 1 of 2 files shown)


marclop:
Once we merge this PR, it needs to be back ported to the `1.0` branch

karencfv (approved):
⭐ Thanks for tackling this!

Issue #382, opened 2020-09-15
Release ecctl v1.0.1


marclop:
## Version to be released

A `v1.0.1` is necessary to address the bug surfaced in #376.

## Checklist

- [x] Verify a cloud-sdk-go release with the same version has been made.
- [x] Update the cloud-sdk-go dependency.
- [x] Update `VERSION` environment variable in the Makefile and generate changelog.
- [x] Execute the release.
- [x] Create release branch for documentation.
- [x] Modify the conf.yaml file in the docs repository to add the release branch.
- [x] Run a full doc rebuild.
- [x] Celebrate :tada:

PR #398, opened 2020-10-06
release: (1.0) Add hub.protocol https to script

marclop:
## Description

Adds a `git config --global hub.protocol https` to the `update-brew-tap`
script which was failing with:

    fatal: could not read Username for 'https://github.com': No such
    device or address

## Related Issues

#382


karencfv (approved):
🙇‍♀️

PR #399, opened 2020-10-06
release: Add hub.protocol https to script

marclop:
## Description

Adds a `git config --global hub.protocol https` to the `update-brew-tap`
script which was failing with:

    fatal: could not read Username for 'https://github.com': No such
    device or address

## Related Issues

#382


GBWHATSAAP:
if you dont find the solution then [click here](https://thegbapp.com/) and resolve the issue

karencfv (approved):

PR #405, opened 2020-11-11
cmd: Add new verbose configuration flags

marclop:
## Description

Adds two persistent flags: `--verbose-credentials` and `--verbose-file`
to the root cmd and their matching configuration fields. Additionally,
adds a utility function to construct `api.Config` and unit tests for it.

## Related Issues

Depends on #404 

## Types of Changes


marclop:
Unit tests will fail until the linked PR is merged and this PR is rebased with the changes.

cmd/root.go
```diff
@@ -110,6 +110,8 @@ func init() {
 	RootCmd.PersistentFlags().String("pass", "", "Password to use to authenticate (If empty will look for EC_PASS environment variable)")
 	RootCmd.PersistentFlags().String("api-key", "", "API key to use to authenticate (If empty will look for EC_API_KEY environment variable)")
 	RootCmd.PersistentFlags().Bool("verbose", false, "Enable verbose mode")
+	RootCmd.PersistentFlags().Bool("verbose-credentials", false, "When set, it will not redact the Authorization headers on the request/response trail")
```
karencfv, line 113: Given the nature of this flag I think it might be worth it to be a bit more explicit with this text. How about something like:

```suggestion
	RootCmd.PersistentFlags().Bool("verbose-credentials", false, "When set, Authorization headers on the request/response trail will be displayed as plain text")
```

cmd/root.go
```diff
@@ -110,6 +110,8 @@ func init() {
...
+	RootCmd.PersistentFlags().String("verbose-file", "", "If set, it will write the verbose request/response trail to this file")
```
karencfv, line 114:
```suggestion
	RootCmd.PersistentFlags().String("verbose-file", "", "When set, the verbose request/response trail will be written to the defined file")
```

karencfv (approved):
🙇‍♀️

Issue #437, opened 2021-02-10
Creating a deployment note requires a user id


danielmitterdorfer:
## Readiness Checklist

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I am reporting the issue to the correct repository (for multi-repository projects)

## Expected Behavior

When running:

```
ecctl deployment note create $DEPLOYMENT_ID --comment "some comment here" --region="aws-eu-central-1"
```

I'd expect that a new note as specified by the value of the command line argument `--comment` is created for the deployment `$DEPLOYMENT_ID`.

## Current Behavior

The above command provides the output:

```
deployment note add: 1 error occurred:
	* user id cannot be empty
```

When running the above command with `--verbose`, we see:

```
Using config file: /Users/daniel/.ecctl/config.yaml
==================== Start of Request #1 ====================
GET /api/v1/deployments/DEPLOYMENT_ID_REDACTED?convert_legacy_plans=false&enrich_with_template=true&show_metadata=false&show_plan_defaults=false&show_plan_history=false&show_plan_logs=false&show_plans=false&show_security=false&show_settings=false&show_system_alerts=5 HTTP/1.1
Host: api.elastic-cloud.com
User-Agent: ecctl/v1.1.0
Accept: application/json
Authorization: [REDACTED]
Accept-Encoding: gzip


====================  End of Request #1  ====================
==================== Start of Response #1 ====================
HTTP/1.1 200 OK
Transfer-Encoding: chunked
Connection: keep-alive
Content-Type: application/json
Date: Wed, 10 Feb 2021 07:53:01 GMT
Server: fac/[redacted]
X-Cloud-Resource-Created: 2020-01-24T10:56:55.362Z
X-Cloud-Resource-Last-Modified: 2021-02-01T17:02:34.945Z
X-Cloud-Resource-Version: 10
X-Ratelimit-Interval: 1 second
X-Ratelimit-Limit: 50
X-Ratelimit-Remaining: 49
X-Ratelimit-Reset: Wed, 10 Feb 2021 07:53:01 GMT
X-Request-Id: [REQUEST_ID_REDACTED]

[JSON BODY REDACTED]

====================  End of Response #1  ====================
deployment note add: 1 error occurred:
	* user id cannot be empty
```

## Steps to Reproduce

1. Create a configuration file in `~/.ecctl/config.yaml` (see contents below)
[...]

PR #446, opened 2021-02-17
docs: Set deployment node command as ECE only

karencfv:
## Description
When adding documentation regarding ECE specific commands
we forgot to mark the `ecctl deplyoment note` command and its
subcommands as one of them. This patch fixes this issue.

## Related Issues
Closes https://github.com/elastic/ecctl/issues/437

## Motivation and Context
Let's not confuse our users :)

## How Has This Been Tested?
By running `make docs`

## Types of Changes

- [x] Documentation


marclop (approved):
LGTM. I'm thinking we should also back port to the 1.2 branch

PR #462, opened 2021-03-17
comment: Adjust text to specify the command is ECE only

karencfv:
## Description
Modifies `ecctl comment` command and subcommand's text to
specify these are ECE-only commands.

## Types of Changes

- [x] Documentation


marclop (approved):
I missed that, thanks!

Issue #448, opened 2021-02-17
[request]add custom bundle upload feature


TomonoriSoejima:
Can you give a thought about adding bundle upload feature?

Right now updating bundle is possible for ECE by following this guide

https://www.elastic.co/guide/en/cloud-enterprise/current/ece-add-custom-bundle-plugin.html#ece-add-custom-plugin

But it is not permitted for Elastic Cloud and doing so will result in error below.

```
    "errors": [
        {
            "code": "clusters.invalid_extension_url",
            "message": "Usage of [plugin] with URL [https://xxxx/ok.zip] is not permitted. The URL format is invalid.",
            "fields": [
                "resources.elasticsearch[0]"
            ]
        },
```

It would be nice if we can make `ecctl` to do this.
Right now bundle upload is the only thing that requires manual labor.

TomonoriSoejima:
It looks like there is a REST [API](https://www.elastic.co/guide/en/cloud/current/Deployments_-_Extensions.html#ec_request_example_55) available and I was able to upload a zip file like this using [HTTP](https://httpie.io/docs#file-upload-forms)

```
http -f  PUT https://api.elastic-cloud.com/api/v1/deployments/extensions/2318760402  file@ok.zip \

Authorization:'ApiKey your-api-key' \

ApiKey:'your-api-key'
```

So I could live with that for now.

PR #470, opened 2021-03-23
deployment: New extension create subcommand

karencfv:
## Description
Adds support for create and upload actions for extensions

- `create <extension id> --version <version> --type <extension type> {--file <file-path> | --download-url <url>} [--description <description>]`: Creates an extension.

## Related Issues
Closes: https://github.com/elastic/ecctl/issues/448

## Motivation and Context
Parity with API

## How Has This Been Tested?
Manually and unit tests.

## Types of Changes

- [x] Documentation

## Readiness Checklist


cmd/deployment/extension/create.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/elastic/cloud-sdk-go/pkg/api/deploymentapi/extensionapi"
+	"github.com/spf13/cobra"
+
+	"github.com/elastic/ecctl/pkg/ecctl"
+)
+
+var createCmd = &cobra.Command{
+	Use:     "create <extension id> --version <version> --type <extension type> {--file <file-path> | --download-url <url>} [--description <description>]",
+	Short:   "Creates an extension",
+	PreRunE: cobra.MinimumNArgs(1),
```
marclop, line 33:
```suggestion
	PreRunE: cobra.ExactArgs(1),
```

cmd/deployment/extension/create.go
```diff
@@ -0,0 +1,93 @@
...
+	"os"
+
+	"github.com/elastic/cloud-sdk-go/pkg/api/deploymentapi/extensionapi"
+	"github.com/spf13/cobra"
+
+	"github.com/elastic/ecctl/pkg/ecctl"
+)
+
+var createCmd = &cobra.Command{
+	Use:     "create <extension id> --version <version> --type <extension type> {--file <file-path> | --download-url <url>} [--description <description>]",
```
marclop, line 31:
```suggestion
	Use:     "create <extension name> --version <version> --type <extension type> {--file <file-path> | --download-url <url>} [--description <description>]",
```

cmd/deployment/extension/create.go
```diff
@@ -0,0 +1,93 @@
...
+			res2, err := extensionapi.Upload(extensionapi.UploadParams{
+				API:         ecctl.Get().API,
+				ExtensionID: *res.ID,
+				File:        f,
+			})
+			if err != nil {
+				return err
+			}
+
+			return ecctl.Get().Formatter.Format("", res2)
```
marclop, line 73: If we return here, we'll never print the first response back to the stdout. Maybe we can change this to something like:
```suggestion
			if err:= ecctl.Get().Formatter.Format("", res2); err != nil
				return err
			}
```

cmd/deployment/extension/create.go
```diff
@@ -0,0 +1,93 @@
...
+	},
+}
+
+func init() {
+	initCreateFlags()
+}
+
+func initCreateFlags() {
+	Command.AddCommand(createCmd)
+	createCmd.Flags().String("version", "", "Elastic stack version. Numeric version for plugins, e.g. 2.3.0. Major version e.g. 2.*, or wildcards e.g. * for bundles.")
```
marclop, line 86: I know this was copied over from the API docs, but maybe worth changing it to something like `7.10.0` and `7.*` respectively to be a little bit more up to date.

cmd/deployment/extension/create_test.go
```diff
@@ -0,0 +1,151 @@
...
+							},
+							mock.NewByteBody(createRawResp),
+						),
+					},
+				},
+			},
+			want: testutils.Assertion{
+				Err: string(createJSONOutput) + "\n",
+			},
+			// Not testing for file upload as the content type is multipart/form-data
```
marclop, line 140: We should add the tests case for the file, we don't need to assert the request though, only the response.

cmd/deployment/extension/create.go
karencfv: hmmmm... I copied this from the UI 🤔 I think it was done this way to make it obvious it's behind?

marclop: I think it's just outdated, our docs are [clearly up to date](https://www.elastic.co/guide/en/cloud/current/ec-custom-bundles.html#ec-add-your-plugin).

cmd/deployment/extension/create.go
karencfv: There's no need to print out the first response as they both return a `models.Extension`, but the second response contains the updated information with the uploaded extension. So it would look a bit weird/redundant I think

marclop: ah, ignore the comment then

marclop (approved):

Issue #484, opened 2021-04-09
Support proxies settings operations


ppapapetrou76:
Proxies settings operations as defined [here](https://www.elastic.co/guide/en/cloud-enterprise/current/Platform_-_proxies.html)  are not yet supported

This is a ticket to track the implementation of the following
- [x] Get proxies settings ( https://github.com/elastic/cloud-sdk-go/pull/325 and https://github.com/elastic/ecctl/pull/485 ) 
- [ ] Update proxies settings ( https://github.com/elastic/ecctl/pull/487 , https://github.com/elastic/cloud-sdk-go/pull/326 , https://github.com/elastic/cloud-sdk-go/pull/327 and https://github.com/elastic/cloud-sdk-go/pull/328 )

PR #487, opened 2021-04-12
Implement proxy settings update command

ppapapetrou76:
## Description
Implements a command to update proxies settings. The command accepts a json file and can be used to either do a partial update ( patch ) - default behavior or replace the existing proxies settings

## Related Issues
Closes https://github.com/elastic/ecctl/issues/484

## How Has This Been Tested?
Against ESS Staging using both update flavors ( patch/put )

## Types of Changes
- [ ] Documentation

## Readiness Checklist


ppapapetrou76:
@marclop @karencfv please let me know if my approach is according to the guidelines. I didn't find any other command that supports both a full and a partial update of an entity and I decided to implement only one with a flag. If this is not ok, let me know and I'll split it into two sub-commands.

docs/ecctl_platform_proxy_settings_update.md
```diff
@@ -0,0 +1,45 @@
...
+      --region string         Elasticsearch Service region
+      --timeout duration      Timeout to use on all HTTP calls (default 30s)
+      --trace                 Enables tracing saves the trace to trace-20060102150405
+      --user string           Username to use to authenticate (If empty will look for EC_USER environment variable)
+      --verbose               Enable verbose mode
+      --verbose-credentials   When set, Authorization headers on the request/response trail will be displayed as plain text
+      --verbose-file string   When set, the verbose request/response trail will be written to the defined file
+```
+
+### SEE ALSO
```
alaudazzi, line 42:
```suggestion
### See also
```

docs/ecctl_platform_proxy_settings_update.md
```diff
@@ -0,0 +1,45 @@
+## ecctl platform proxy settings update
+
+Updates proxies settings (Available for ECE only)
```
alaudazzi, line 3:
```suggestion
Updates proxy settings (Available for ECE only)
```

alaudazzi (commented):
A couple of minor suggestions:
- The singular form of proxy seems more readable. 
- Initial cap for headings. SEE ALSO => See also.

docs/ecctl_platform_proxy_settings_update.md
ppapapetrou76: @alaudazzi this is auto-generated 🤷🏾  - I can't modify it.

docs/ecctl_platform_proxy_settings_update.md
ppapapetrou76: @alaudazzi this operation changes the settings for all proxies, not a single one. Do we still want to use the singular form?

docs/ecctl_platform_proxy_settings_update.md
karencfv: We had a conversation with @nrichers about this too 😄 But yeah, what @ppapapetrou76 said, we can't change it.

docs/ecctl_platform_proxy_settings_update.md
karencfv: What about 
```suggestion
Updates settings for all proxies (Available for ECE only)
```

cmd/platform/proxy/settings/update.go
```diff
@@ -0,0 +1,94 @@
...
+	initUpdateFlags()
+}
+
+func initUpdateFlags() {
+	Command.AddCommand(platformProxySettingsUpdateCmd)
+	platformProxySettingsUpdateCmd.Flags().String("file", "", "File name containing the proxy settings")
+	platformProxySettingsUpdateCmd.Flags().String("version", "", "If specified, checks for conflicts against the version of the settings")
+	platformProxySettingsUpdateCmd.Flags().Bool("patch", true, "Set this to false to perform a full update and "+
+		"overwrite the proxy settings - all unspecified fields are deleted.\n"+
+		"When it's set to true, the operation does a partial update and only the fields that are referenced in the given file are changed")
```
karencfv, line 87: Generally when we have boolean flags, they are set to false, so I think it's probably a good idea to do this de other way around. Have the flag be `--full` instead of patch and set it to false.
```suggestion
	platformProxySettingsUpdateCmd.Flags().Bool("full", false, "If set, a full update will be performed and all proxy settings will be overwritten. Any unspecified fields will be deleted.")
```

cmd/platform/proxy/settings/update.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/spf13/cobra"
+
+	cmdutil "github.com/elastic/ecctl/cmd/util"
+	"github.com/elastic/ecctl/pkg/ecctl"
+)
+
+var platformProxySettingsUpdateCmd = &cobra.Command{
+	Use:     "update --file settings.json",
+	Short:   cmdutil.AdminReqDescription("Updates proxies settings"),
+	PreRunE: cobra.ExactArgs(0),
```
karencfv, line 36: Given the complexity if this command, it's probably a good idea to have an `Example:` field with some examples :)

karencfv (commented):
Thanks a bunch for implementing this ❤️. I've left a few comments

ppapapetrou76:
@karencfv thanks a lot for the detailed review ! I think I addressed all comments

cmd/platform/proxy/settings/update.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/elastic/cloud-sdk-go/pkg/models"
+	sdkcmdutil "github.com/elastic/cloud-sdk-go/pkg/util/cmdutil"
+	"github.com/spf13/cobra"
+
+	cmdutil "github.com/elastic/ecctl/cmd/util"
+	"github.com/elastic/ecctl/pkg/ecctl"
+)
+
+const examples = `## Update only the defined proxy settings 
+$ ecctl platform proxy settings update -file settings.json --region=us-east-1
```
karencfv, line 34:
```suggestion
$ ecctl platform proxy settings update --file settings.json --region us-east-1
```

cmd/platform/proxy/settings/update.go
```diff
@@ -0,0 +1,99 @@
...
+
+## Update by overriding all proxy settings 
+$ ecctl platform proxy settings update --file settings.json --region=us-east-1 --full
```
karencfv, line 37:
```suggestion
$ ecctl platform proxy settings update --file settings.json --region us-east-1 --full
```

karencfv (approved):
Thanks for addressing my comments! 🙇‍♀️

PR #497, opened 2021-08-18
release: Update cloud-sdk-go to v1.5.0

marclop:
## Description

Updates the elastic/cloud-sdk-go dependency to `v1.5.0`. Also disabling the new `completions` command that was added in the new version of `spf13/cobra`.


marclop:
the Cobra dependency added a "completions" command to the root cmd, which is breaking the doc generation. I will disable that default cmd for now and open an issue so we can investigate in due time.

artemnikitin (approved):

obierlaire (approved):

PR #501, opened 2021-08-19
release: Update releasing doc

marclop:
## Description

Updates the `RELEASE.md` doc with the latest information on how to ship
a new version of `ecctl`.

### DIFF
--- a/developer_docs/RELEASE.md
+++ b/developer_docs/RELEASE.md
@@ -1,6 +1,17 @@
 # Releasing a new version
 
-This guide aims to provide guidance on how to release new versions of the `ecctl` binary as well as updating all the necessary parts to make it successful.
+This guide aims to provide guidance on how to release new versions of the `ecctl` binary as well as updating all the necessary parts to make it successful. The release will happen automatically via GitHub actions, but there are a few prerequisites to tick before that can be started.
+
+- [Releasing a new version](#releasing-a-new-version)
+  - [Prerequisites](#prerequisites)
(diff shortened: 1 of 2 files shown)


developer_docs/RELEASE.md
```diff
@@ -25,30 +36,36 @@ SHELL := /bin/bash
...
 
-Say we want to perform a minor version release (i.e. no breaking changes and only new features and bug fixes are being included); in which case we'll update the _MINOR_ part of the version:
+Say we want to perform a minor version release (i.e. no breaking changes and only new features and bug fixes are being included); in which case we'll update the _MINOR_ part of the version, this can be done with the `make minor` target, but it should have been updated automatically via GitHub actions.
 
 ```Makefile
 SHELL := /bin/bash
 export VERSION ?= v1.1.0
 ```
 
+If a patch version needs to be released, the release will be done from the minor branch. For example, if we want to release `v1.5.1`, we will check out the `1.5` branch and perform any changes in that branch. The VERSION variable in the Makefile should already be up to date, but in case it's not, it can be bumped with the `make patch` target.
```
jeredding, line 46: you're using the term `patch` here when above in the semversion definition you have `PATCH` (excerpt: `in the SEMVER format: MAJOR.MINOR.BUG`) - can we make consistent?

jeredding (approved):
LGTM aside from one edit to address.  

I was curious - `make tag` does that require any command line args?

developer_docs/RELEASE.md
marclop: you're right, I'll fix that.

PR #506, opened 2021-08-24
init: Add an ESS API Key creation link

### LINKED ISSUE #505: [discuss] Configuration and API Key documentation
## Question

The [Configuring page](https://www.elastic.co/guide/en/ecctl/current/ecctl-configuring.html) talks about the precedence of configuration sources, and how to generate the config file, but doesn't explain where to get the API key that is needed for `ecctl init` ([this page](https://www.elastic.co/guide/en/cloud/current/ec-api-authentication.html) for ESS and [this section](https://www.elastic.co/guide/en/cloud-enterprise/current/ece-restful-api-authentication.html#ece-api-keys) for ECE). It might be good to mention that above the `ecctl init` steps. For ESS we can link directly to https://cloud.elastic.co/deployment-features/keys .

I also think we'd prefer to de-emphasize the `--user` and `--pass` options (which I don't think work at all for ESS). But I didn't want to make those changes without discussing it first.

marclop:
## Description

Adds a link in the `ecctl init` workflow when the Elasticsearch Service
option has been selected to the ESS API Key feature page.
Additionally, updates the configuration precedence documentation to use
a different set of flags rather than `--user` and `--pass`, since these
are only relevant for ECE users.

## Related Issues

Closes #505 

## How Has This Been Tested?

Unit & manual tested.

## Types of Changes

- [x] Documentation

### DIFF
--- a/docs/ecctl-getting-started.asciidoc
+++ b/docs/ecctl-getting-started.asciidoc
@@ -108,7 +108,7 @@ variables, or at runtime using the CLI’s global flags.
 === Before you begin
 The hierarchy for configuration parameters is as follows, from higher precedence to lower:
 
-. Command line flags `--host`, `--user`, `--pass`
+. Command line flags `--api-key`, `--region`, `--verbose`
 . Environment variables
 . Shared configuration file
 `$HOME/.ecctl/config.<json|toml|yaml|hcl>`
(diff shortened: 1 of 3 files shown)


gtback (approved):
LGTM! I haven't tested it manually, though

rhass (approved):
LGTM

PR #526, opened 2021-09-29
Fix env variable name with Github token

artemnikitin:
## Description

This is a follow-up fix for https://github.com/elastic/ecctl/pull/524. In the added CI job name of the env variable with the Github token was set incorrectly.


rhass (approved):
LGTM

PR #534, opened 2021-12-01
docs: Add note about the EC Terraform provider

marclop:
## Description

Mentions the EC Terraform provider as the preferred tool to manage
deployments.

## Related Issues

None

## Motivation and Context

Some users are still relying on `ecctl` to create and manage ESS deployments, while the tool allows it since it can create deployments from a JSON definition, it's not the most suitable tool for the job.

## Types of Changes

- [x] Documentation


artemnikitin (approved):

artemnikitin:
Thank you, Marc 👍

PR #649, opened 2024-05-15
Expose clear transients in ecctl

nick-benoit:
## Description

Exposes new `clear_transients` field for ecctl for show deployment. 

The default value depends on on the value of `generate-update-payload` in order to facilitate safe usage of the "read - edit - write" loop. If `generate-update-payload=true` then `clear-transients` also defaults to true and is false otherwise. If the flag is manually passed then that value takes precedence.

## Related Issues

https://elasticco.atlassian.net/browse/CP-4726

## Motivation and Context

This change makes reading and writing deployments safer. We ran into this during the last security remediation and almost overwrote someone’s cluster from their most recent snapsho

## How Has This Been Tested?

This has been tested by using local build to operate against QA ESS. I have verified that with 
`--generate-update-payload=true` the transient properties are not included (default behavior). If `--generate-update-payload=true` and `--clear-transient=false` is explicitly passed then transient properties are included. 

## Types of Changes

- [ ] Documentation

## Readiness Checklist

- [x] have added tests to cover my changes


cmd/deployment/show.go
```diff
@@ -112,4 +123,5 @@ func initShowFlags() {
 	showCmd.Flags().BoolP("metadata", "m", false, "Shows the deployment metadata")
 	showCmd.Flags().BoolP("settings", "s", false, "Shows the deployment settings")
 	showCmd.Flags().Bool("generate-update-payload", false, "Outputs JSON which can be used as an argument for the --file flag with the update command.")
+	showCmd.Flags().Bool("clear-transient", false, "Removes the `transient` field in order to make read - edit - write loop safe")
```
AlexP-Elastic, line 126: should the comment explain that its default is true when generate-update-payload is enabled?

AlexP-Elastic (commented):
functionally LGTM though someone more familiar (Dominik?) should give it the "binding review" :)

cmd/deployment/show.go
nick-benoit: Yeah I think so. I updated documentation.

docs/ecctl_deployment_show.adoc
```diff
@@ -26,16 +26,17 @@ ecctl deployment show <deployment-id> [flags]
...
-  -h, --help                      help for show
-      --kind string               Optional deployment resource kind (apm, appsearch, kibana, elasticsearch)
-  -m, --metadata                  Shows the deployment metadata
-      --plan-defaults             Shows the deployment plan defaults
-      --plan-history              Shows the deployment plan history
-      --plan-logs                 Shows the deployment plan logs
-      --plans                     Shows the deployment plans
-      --ref-id string             Optional deployment kind RefId, if not set, the RefId will be auto-discovered
-  -s, --settings                  Shows the deployment settings
+      --clear-transient transient   Removes the transient field in order to make read - edit - write loop safer. The default value of clear-transient depends on the value of `generate-update-payload`. If `generate-update-payload` is true `clear-transient` defaults to true. Otherwise defaults to false.
```
gigerdo, line 29: Looks like some issue with the docs generator: It adds a transient that shouldn't be there `--clear-transient transient    Removes the`

It seems like it is due to `` in the text. Maybe some special syntax 🤔

gigerdo (approved):
LGTM, just two comments on a small issue with the docs and maybe a test that could be added

PR #664, opened 2024-09-19
Deployment search: Add flag to return all matches.

gigerdo:
- By default the deployment search only makes one request, returning the number of results specified in 'size'
  - This only works up to 10k results, but may also already stop working before that just due to the size of the response.
- A new flag now switches to using a cursor to request all matches
- This works by making a request, then passing the returned cursor into the next request for the next batch of results

## Motivation and Context

Being able to search >10k deployments.

## Types of Changes

- [ ] Documentation

## Readiness Checklist


cmd/deployment/search.go
```diff
@@ -51,21 +54,57 @@ var searchCmd = &cobra.Command{
...
+		returnAllMatches, _ := strconv.ParseBool(cmd.Flag("all-matches").Value.String())
+		if returnAllMatches && sr.Sort == nil {
+			return fmt.Errorf("The query must include a sort-field when using --all-matches. Example: \"sort\": [\"id\"]")
+		}
 
-		if err != nil {
-			return err
+		var result *models.DeploymentsSearchResponse
+		var cursor string
+		for {
```
nick-benoit, line 64: This is probably just me being overly paranoid, but does it make sense to add some kind of failsafe that kills the loop after a certain (high) number of requests? 

That way a future bug in the AC cursor logic or this tool doesn't end up in an infinite tight loop making search requests?

nick-benoit (approved):
LGTM 🚀  - As a side note for some reason the cursor searching doesn't appear in the [API docs.](https://www.elastic.co/guide/en/cloud-enterprise/current/search-deployments.html)

cmd/deployment/search.go
gigerdo: Yeah it could make sense to have some timeout/upper limit

cmd/deployment/search.go
```diff
@@ -51,21 +54,57 @@ var searchCmd = &cobra.Command{
...
+			sr.Cursor = cursor
+			if returnAllMatches {
+				// Custom batch-size to override any size already set in the input query
+				sr.Size = 500
```
dimuon, line 68: Does it make sense to read it from cmd arg as well? (while "500" can be the default value).

dimuon (approved):

cmd/deployment/search.go
gigerdo: Yeah it could be useful if this size turns out to be wrong in some situations

gigerdo:
> As a side note for some reason the cursor searching doesn't appear in the [API docs.](https://www.elastic.co/guide/en/cloud-enterprise/current/search-deployments.html)

It doesn't show up because 3.7 doesn't support it. 3.8 will support it, but is not released yet. So currently only the cloud API will show it: https://www.elastic.co/guide/en/cloud/current/Deployment_-_Commands.html#search-deployments

cmd/deployment/search.go
gigerdo: I added a --size parameter

PR #666, opened 2024-09-26
Prepare 1.14.0 release

gigerdo:
- Update version to 1.14.0
- Add release notes
- Update notice


NOTICE
```diff
@@ -71,10 +71,10 @@ THE SOFTWARE.
 --------------------------------------------------------------------------------
 
 Dependency : github.com/elastic/cloud-sdk-go
-Version: v1.16.0
+Version: v1.22.0
```
claudia-correia, line 74: In the meantime, v1.23.0 of the SDK was released, not sure if it's worth it to update this 🙂

claudia-correia (approved):
LGTM 👍

NOTICE
gigerdo: I think I'll leave that for when we have changes that need it (Since I think it would have no effect on ecctl to update that unless we use any new API functionality)

PR #667, opened 2024-09-27
Fix goreleaser-action

gigerdo:
The release github action is currently broken:

We were using the `latest` gorelease-action, which had changed from 1->2 recently. The main breaking change for us was that `--rm-dist` flag was replaced by `--clean`

But this change causes some other changes:
- Some smaller changes in the goreleaser config
- To run the newest goreleaser, we have to upgrade to golang 1.23.0
- When using go 1.23.0, golangci-lint also needs upgrading
- golanci-lint on the new version also has some config deprecations/changes 

But with all these changes, the local build/lint runs without errors or warnings and the dry-run gorelease also runs through (hopefully the same is the case in the github action).

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.20"
+          go-version: "1.23"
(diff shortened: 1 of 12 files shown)


claudia-correia (approved):
👍

PR #711, opened 2025-03-20
Update branch.yml

ismisepaul:
Pinning action to a full length commit SHA [see](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)


gigerdo (approved):
Thanks for the improvement. LGTM 👍

PR #733, opened 2025-07-03
Fix TOC so docs build can include ECCTL release notes

shainaraskas:
related to https://github.com/elastic/docs-builder/issues/764

PR #734, opened 2025-07-03
add a hidden file to make the child TOC valid

shainaraskas:
Redoing my fix from https://github.com/elastic/ecctl/pull/733: the fix wasn't valid for the location where the release notes need to be hosted


bmorelli25 (approved):
looks right to me

nick-benoit (approved):

